PR #248, opened 2023-04-03
Publish to NPM

### LINKED ISSUE #246: [FEATURE] Publish to NPM
## User Story
1. As a {Jayvee user}
2. I want {to use NPM as package registry}
3. So that {I don't need to get a GitHub PAT first when installing Jayvee}

## User Acceptance Criteria
- [ ] Packages are published to NPM

## Definitions of Done
- [ ] All tests are passing

georg-schwarz:
Closes #246.

This PR is more complex than I wished for. The following things were necessary to make it work:
- Use different repository in CI and `package.json` files
- Use better names for the NPM packages (appended `jayvee-` for most packages)
- Adjust the local alias paths in the `tsconfig.base.json` to the package names (otherwise build fails), adjust all imports
- Remove the `@jvalue` prefix of the `language-server` and the `monaco-editor` in their `project.json` to make that consistent with the other projects, adjust the use of the nx/npm commands accordingly 

All in all, that PR is very prone to errors. Locally, I tested by running all examples, and generating the docs, and installing the vscode plugin. I did not manually test the web worker and the monaco editor.

After merging, the CI is to be evaluated if the NPM publishing works. Not tested manually as it would lead to version conflicts on publish via CI.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -24,7 +24,7 @@ jobs:
           registry-url: 'https://registry.npmjs.org'
       - run: npm ci
       - name: Publish language server
-        run: npx nx run @jvalue/language-server:publish
+        run: npx nx run language-server:publish
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
       - name: Publish interpreter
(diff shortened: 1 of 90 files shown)


rhazn (approved):
Hmm, I'll have a hard time reasonably reviewing this because I am not sure what the reasoning behind adding @jvalue to some of the names was... Overall there is nothing obviously wrong so we can try this.

For lower impact, I'd maybe just do the interpreter publishing to npm right now and keep the rest as it is, that should be possible as well, right?

georg-schwarz:
> For lower impact, I'd maybe just do the interpreter publishing to npm right now and keep the rest as it is, that should be possible as well, right?

Sure, the "unnecessary" changes are encapsulated in commit b68926bf95b684a2eacccdb874eb5a8c4ab17e33. I can move that one to another PR.

rhazn:
I think that would be the best setup, do the minimal needed changes for a public interpreter now and wait for felix for the rest.

PR #249, opened 2023-04-04
Make nx project names consistent

georg-schwarz:
Did we have a good reason to have the appendix `@jvalue` for some of the projects and for some not? I assume it is just some technical dept that we can clean up?


felix-oq (approved):
I think the `@jvalue` prefix was always autogenerated by Nx for libraries.

Overall I like the change but I'm not sure whether Nx will still add the prefix when generating new libraries.

PR #250, opened 2023-04-04
Installation docs

georg-schwarz:
## GitHub Project Page
Added an image to link to the docs. It should be obvious now to click it to get further info

image

Mobile:

## Docs
Now contains installation instructions for interpreter and VSCode plugin.

image


georg-schwarz:
Will go ahead, no need for a throughout review from my side. If you still should catch something I'll make a follow-up PR to fix/adapt.

Issue #219, opened 2023-03-20
[FEATURE] GTFS-RT Support (RFC0006)


schlingling:
## User Story
1. As a {user}
2. I want {to archive GTFS-RT-data from a http-endpoint}
3. So that {multiple executions of a pipeline containing both, GTFS and GTFS-RT sections, demonstrate an archiving process of static as well as realtime GTFS data}

## Notes
A basis for all UACs is the corresponding [RFC0006 GTFS-RT Support](https://github.com/jvalue/jayvee/tree/main/rfc/0006-gtfs-rt-support). The RFC got qualified by multiple iterations, which are listed below. One UACs represents a single requirement, extraced from the final, accepted iteration.
Iterations of [RFC0006 GTFS-RT Support](https://github.com/jvalue/jayvee/tree/main/rfc/0006-gtfs-rt-support):
| **Iteration** | **Scope**                                                                 | **PR** |
|---------------|---------------------------------------------------------------------------|--------|
| 1             | GTFS-RT file processing using GtfsRTInterpreter: inital concept                  | #200   |
| 2             | GTFS-RT file processing using GtfsRTInterpreter: refinement after feedback-loop   | #201   |

## User Acceptance Criteria
- [x] [UAC-1] New blocktype `GtfsRTInterpreter` is implemented in std-extension. via #223 
- [x] [UAC-1.1] A new demo pipeline `gtfs-rt-simple.jv` is implemented. via #223
- [x] [UAC-2] The current blocktype `SQLiteSink` is configurable by an attribute `dropTable` indicating to drop data before loading to the sink. iva #254
- [x] [UAC-3] `gtfs-static-and-rt.jv` is added to showcase the processing of GTFS-RT data as well as GTFS data. via #255
- [x] [UAC-4] Every run of `gtfs-static-and-rt.jv` creates/updates ONE sqlite file via #255
- [x] [UAC-5] Every run of `gtfs-static-and-rt.jv` downloads GTFS data and overwrites GTFS tables via #255
- [x] [UAC-6] Every run of `gtfs-static-and-rt.jv` downloads GTFS-RT data and appends it to GTFS-RT tables via #255
- [x] [UAC-7] Jayvee processes successfully `gtfs-static-and-rt.jv` via #255

## Examples
A detailled explanation of all UACs and further context is provided by [RFC0006 GTFS-RT Support](https://github.com/jvalue/jayvee/tree/main/rfc/0006-gtfs-rt-support)

## Definitions of Done
[...]

PR #254, opened 2023-04-08
The current blocktype SQLiteSink is configurable by an attribute dropTable indicating to drop data before loading to the sink (UAC-2 of #219)

schlingling:
The current blocktype SQLiteSink is configurable by an attribute dropTable indicating to drop data before loading to the sink. DropTable-attribute defaults to true. Appending GTFS-RT-Data periodically works now (see pipeline `gtfs-rt-simple.jv`)

Uses package [gtfs-realtime-binding](https://github.com/MobilityData/gtfs-realtime-bindings/tree/master) published under Apache2.0 license

Part of [RFC0006 GTFS-RT Support](https://github.com/jvalue/jayvee/tree/main/rfc/0006-gtfs-rt-support).

Implements following UACs of #219:
- [x] [UAC-2] The current blocktype `SQLiteSink` is configurable by an attribute `dropTable` indicating to drop data before loading to the sink.

Please merge first #223 to main, before merging that one.


rhazn (changes requested):
Looks okay but why is this PR to main, it should be to the branch of #223 , no? Please rebase this so it only contains the changes that are actually mentioned in the description.

schlingling:
> Looks okay but why is this PR to main, it should be to the branch of #223 , no? Please rebase this so it only contains the changes that are actually mentioned in the description.

Yes no problem. Thought to merge it directly to main in assumption that #223 is then already in main.

rhazn (approved):

PR #256, opened 2023-04-10
Implemented composite pattern for filesystem

### LINKED ISSUE #131: Refactor FileSystem to Compositum
This is fine and imho we can merge it like this. In a separate MR I'd encourage you to move this to a composite-pattern implementation, with directories and files (see this eloquent example by a well known german professor https://www.youtube.com/watch?v=efDxnGi8zU8&t=60s).

_Originally posted by @rhazn in https://github.com/jvalue/jayvee/pull/126#discussion_r1101557677_

Actually the FileSystem follows already a composite approach, but could be more explicit.

schlingling:
closes #131

Introduces a clean composite pattern for all filesystem related artefacts like it is done in ADAP.

image

The naming convention of the .ts-files yields the class-hierachy, so it is easier to understand.

image


schlingling:
@rhazn FYI, since you are on leave, i requested Felix for an Review :)

rhazn:
Thank you, if this is still open next week ping me and I can take a look then as well. If Felix is fine with it that's enough as well of course.

libs/execution/src/lib/types/io-types/filesystem.ts
felix-oq, line 1: An idea I have in mind: Introduce a generic type for FileSystem that is used as a type for files. Something like `FileSystem<F extends FileSystemFile>` and in the application probably use it as `FileSystem<BinaryFile>`. What do you think about that?

libs/execution/src/lib/types/io-types/filesystem-inmemory.ts
```diff
@@ -4,94 +4,96 @@
...
-      // Check, if directory already exists, if not create it
-      let childFileSystemIndex = currentFileSystemIndex.get(part) as
-        | InMemoryFileSystem
-        | undefined;
-      if (!childFileSystemIndex) {
-        childFileSystemIndex = new InMemoryFileSystem();
-        currentFileSystemIndex.set(part, childFileSystemIndex);
+  private processPath(path: string): string[] {
+    const parts = path
+      .toLowerCase()
```
felix-oq, line 83: Why do you use `toLowerCase` here? That way, users cannot refer to files / folders that have capital letters in their names.

libs/execution/src/lib/types/io-types/filesystem-inmemory.ts
```diff
@@ -4,94 +4,96 @@
...
+      .split(InMemoryFileSystem.PATH_SEPARATOR)
+      .filter((p) => p !== ''); // Process paths like "folder1//folder1" to "folder1/folder2"
+    const processedParts: string[] = [];
+    for (const part of parts) {
+      if (part === InMemoryFileSystem.CURRENT_DIR) {
+        continue; // Skip current dirs in path
+      }
+      if (part === InMemoryFileSystem.PARENT_DIR) {
+        processedParts.pop(); // Go level up in folder hierarchy if ..
```
felix-oq, line 92: What if `processedParts` is empty at this point? In your current implementation, a path `../folder/file.txt` is semantically equal to `./folder/file.txt` which should not be the case.

libs/execution/src/lib/types/io-types/filesystem-node-directory.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  override getNodeSize(): number {
+    return this.children.reduce(
+      (totalSize, child) => totalSize + child.getNodeSize(),
+      0,
+    );
+  }
+
+  add(fileSystemNode: FileSystemNode) {
+    this.children.push(fileSystemNode);
```
felix-oq, line 21: You need to pay attention to duplicate names here. E.g. if a user tries to add a file `file.txt` but a file with the same name already exists as a child of this directory.

libs/execution/src/lib/types/io-types/filesystem-node-directory.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  remove(fileSystemNode: FileSystemNode): FileSystemNode {
+    const index = this.children.indexOf(fileSystemNode);
+    if (index !== -1) {
+      this.children.splice(index, 1);
+    }
+    return this;
+  }
+
+  findFile(fileName: string): FileSystemNode | undefined {
```
felix-oq, line 32: I'd expect a `FileSystemFile` as a return value here.

Also, you should shift this method to `FileSystemNode` and have the different implementations in its subclasses. That way you can avoid `instanceof` checks.

And also note that this method returns a file with the given name that is found first. To find a specific file at a given path, you need the path as a parameter, so it can be descent recursively along the path.

libs/execution/src/lib/types/io-types/filesystem-node-directory.ts
```diff
@@ -0,0 +1,52 @@
...
+        const file = child.findFile(fileName);
+        return file ?? undefined;
+      }
+    }
+    return undefined;
+  }
+
+  getChildNode(childName: string): FileSystemNode | undefined {
+    for (const child of this.children) {
+      if (child.name.toLowerCase() === childName) {
```
felix-oq, line 46: Also here, I think you should not apply `toLowerCase` to any names so we can distinguish between lower case and capital letters.

libs/execution/src/lib/types/io-types/filesystem-inmemory.ts
```diff
@@ -4,94 +4,96 @@
...
   public readonly ioType = IOType.FILE_SYSTEM;
 
+  private rootDirectory: FileSystemDirectory = new FileSystemDirectory('root');
   private static PATH_SEPARATOR = '/';
   private static CURRENT_DIR = '.';
   private static PARENT_DIR = '..';
 
-  // Hierachical file system
-  private fileSystemIndex: Map<string, BinaryFile | FileSystem> = new Map();
+  getFile(path: string): FileSystemFile<unknown> | null {
```
felix-oq, line 20: After refactoring the `findFile` method, it can be called here on `rootDirectory` to find the desired file.

libs/execution/src/lib/types/io-types/filesystem-inmemory.ts
```diff
@@ -4,94 +4,96 @@
...
+    if (child !== undefined && child instanceof FileSystemFile) {
+      return child;
+    }
+    return null;
   }
 
-  getFile(filePath: string): BinaryFile | null {
-    const processedParts = this.processPath(filePath);
+  putFile(path: string, file: BinaryFile): FileSystem | null {
+    const processedParts = this.processPath(path);
```
felix-oq, line 47: Similar to `findFile`, you could add a method to `FileSystemNode` that covers this functionality. Then, you could simply call `rootDirectory.putFile(path, file)` (or something similar) at this point to descent recursively and add the file at the desired location. This again would avoid the `instanceof` checks.

felix-oq (changes requested):
To avoid the `instanceof` checks, consider introducing new methods in `FileSystemNode` that perform the recursive descents. Those can then be called in the `FileSystem` class.

Also I think that paths should generally be treated case-sensitive.

libs/execution/src/lib/types/io-types/filesystem.ts
schlingling: Limits it to just store File of the same type in the FileSystem but state now, this functionality is enough. So i like that.

libs/execution/src/lib/types/io-types/filesystem.ts
```diff
@@ -15,17 +16,17 @@ export interface FileSystem extends IOTypeImplementation<IOType.FILE_SYSTEM> {
...
-   * @param {string} filePath - The file path.
-   * @param {File} file - The file to save.
-   * @returns {FileSystem}
+   * @function putNode
+   * @param {string} path - The path to the node.
+   * @param {FileSystemNode} node - The node to save.
+   * @returns {FileSystem | null} - The directory or null if the file failed to insert
    */
-  putFile(filePath: string, file: BinaryFile): FileSystem;
+  putFile(path: string, file: BinaryFile): FileSystem | null;
```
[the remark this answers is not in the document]
schlingling: thanks, changed

libs/execution/src/lib/types/io-types/filesystem.ts
```diff
@@ -15,17 +16,17 @@ export interface FileSystem extends IOTypeImplementation<IOType.FILE_SYSTEM> {
...
   /**
    * Saves a file to the file system.
-   * @function putFile
-   * @param {string} filePath - The file path.
-   * @param {File} file - The file to save.
-   * @returns {FileSystem}
+   * @function putNode
+   * @param {string} path - The path to the node.
+   * @param {FileSystemNode} node - The node to save.
+   * @returns {FileSystem | null} - The directory or null if the file failed to insert
```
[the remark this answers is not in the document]
schlingling: thanks, changed.

libs/execution/src/lib/types/io-types/filesystem-inmemory.ts
schlingling: @rhazn  once raised the requirement, to handle paths case insensitive, thats why i am using toLowerCase() (see https://github.com/jvalue/jayvee/pull/126#discussion_r1098412123)

libs/execution/src/lib/types/io-types/filesystem-node-directory.ts
schlingling: i removed that method since it wasnt used.

libs/execution/src/lib/types/io-types/filesystem-inmemory.ts
schlingling: removed find file, so thats obsolet.

libs/execution/src/lib/types/io-types/filesystem-inmemory.ts
schlingling: My approach was to keep the FileSystemNode (composite) simple for type saftey (just add, and get methods) and move all the logic to the client in this case the `FileSystem`. 

See [wiki article](https://en.wikipedia.org/wiki/Composite_pattern#:~:text=Design%20for%20type%20safety%3A%20Child%2Drelated%20operations%20are%20defined%20only%20in%20the%20Composite%20class.%20Clients%20must%20treat%20Leaf%20and%20Composite%20objects%20differently.%20But%20type%20safety%20is%20gained%20because%20clients%20cannot%20perform%20child%2Drelated%20operations%20on%20Leaf%20objects.), which is describing design for type saftey

libs/execution/src/lib/types/io-types/filesystem-inmemory.ts
rhazn: Hmm I thought that is how unix does it but apparently I was wrong, we can change to care for capitalization, yeah.

schlingling: okay, changed to case sensitive pathing

libs/execution/src/lib/types/io-types/filesystem-node-directory.ts
schlingling: removed it and changed behavior to support case sensitive paths

libs/execution/src/lib/types/io-types/filesystem-node-directory.ts
schlingling: added check if node name exists. This implementation requires now in each node to have unique node names.

Drawback is, that we cant have a dir and a file with same name in same directory, but this can be implemented later on if needed.

libs/execution/src/lib/types/io-types/filesystem-inmemory.ts
schlingling: Ouh thats a good point, thanks for that hint! 

Actually in the current implementation of `InMemoryFileSystem` which always starts at its "virtual" root-dir it does not make sense to call getFile/putFile in general with a relative path (like relativ to what since it is in memory?).

I would mention this in the documentation, if this is okay?

schlingling:
> To avoid the `instanceof` checks, consider introducing new methods in `FileSystemNode` that perform the recursive descents. Those can then be called in the `FileSystem` class.
> 
> Also I think that paths should generally be treated case-sensitive.

Appreciate your feedback, thank you! 
- I changed the behavior of caseinsensitivitiy to casesensitive paths.
- I optimized current implementations in `FileSystemInMemory` simplified code in whole composite hierachy and removed unused peaces
- I followed the [composite pattern variant proposed by wikipedia](https://en.wikipedia.org/wiki/Composite_pattern#:~:text=Design%20for%20type%20safety%3A%20Child%2Drelated%20operations%20are%20defined%20only%20in%20the%20Composite%20class.%20Clients%20must%20treat%20Leaf%20and%20Composite%20objects%20differently.%20But%20type%20safety%20is%20gained%20because%20clients%20cannot%20perform%20child%2Drelated%20operations%20on%20Leaf%20objects.) namley "Design for type saftey" so i need to treat composites differently in the client (FileSystem)

libs/execution/src/lib/types/io-types/filesystem-inmemory.ts
felix-oq: I'd argue for the uniformity approach rather than the type safety approach, since this is emphasized by the composite pattern and leads to more cohesive code in my opinion. Also see [here](https://en.wikipedia.org/wiki/Composite_pattern#:~:text=The%20Composite%20design%20pattern%20emphasizes%20uniformity%20over%20type%20safety.) in the wiki article.

libs/execution/src/lib/types/io-types/filesystem-inmemory.ts
felix-oq: Yes, I'd argue that you should always call it with absolute paths, so you always start in the root directory. Then you should ensure, that the method is only called with absolute paths and error on any paths that ascent beyond the root directory (e.g. `/folder/../..`).

libs/execution/src/lib/types/io-types/filesystem-node-directory.ts
felix-oq: I think this should work right away, so the in-memory file system is semantically equivalent to the one of the corresponding ZIP archive. In case we postpone it, we should definitely track it via an issue.

libs/execution/src/lib/types/io-types/filesystem.ts
```diff
@@ -13,19 +13,19 @@ import { IOTypeImplementation } from './io-type-implementation';
  */
 export interface FileSystem extends IOTypeImplementation<IOType.FILE_SYSTEM> {
   /**
-   * Retrieves a file from the file system.
+   * Retrieves a file from the file system. Depending on the implementation relative Paths (./) and parent directories (../) are treated differntly.
```
felix-oq, line 16: Why might `.` and `..` be treated differently in different implementations? I think they should have fixed semantics (i.e. `.` for current folder, `..` for parent folder), regardless of the concrete implementation.

libs/execution/src/lib/types/io-types/filesystem.ts
```diff
@@ -13,19 +13,19 @@ import { IOTypeImplementation } from './io-type-implementation';
...
+  getFile(path: string): FileSystemFile<unknown> | null;
 
   /**
-   * Saves a file to the file system.
-   * @function putFile
-   * @param {string} filePath - The file path.
-   * @param {File} file - The file to save.
-   * @returns {FileSystem}
+   * Saves a file to the file system. Depending on the implementation relative Paths (./) and parent directories (../) are treated differntly.
+   * @function putNode
```
felix-oq, line 25:
```suggestion
   * @function putFile
```

Or omit the line entirely, because the function name is obvious from the function signature below.

libs/extensions/std/exec/src/archive-interpreter-executor.ts
```diff
@@ -83,7 +82,12 @@ export class ArchiveInterpreterExecutor
             mimeType,
             content,
           );
-          root.putFile(relPath, file);
+          if (root.putFile(relPath, file) == null) {
+            return R.err({
+              message: `File with name ${file.name} and path ${relPath} failed to insert`,
+              diagnostic: { node: context.getCurrentNode(), property: 'name' },
+            });
+          }
```
felix-oq, line 90: Better use `assert` here, because we expect that all `putFile` calls succeed at runtime:
```suggestion
          const addedFile = root.putFile(relPath, file);
          assert(addedFile != null);
```

libs/extensions/std/exec/src/file-picker-executor.ts
```diff
@@ -2,6 +2,8 @@
 //
 // SPDX-License-Identifier: AGPL-3.0-only
 
+import assert = require('assert');
```
felix-oq, line 5:
```suggestion
import {strict as assert} from 'assert';
```

libs/execution/src/lib/types/io-types/filesystem-inmemory.ts
schlingling: is implemented in new version

libs/execution/src/lib/types/io-types/filesystem-inmemory.ts
schlingling: Oh yes you are right, convinced me :D I changed with new commit to uniformity approach using clean recursion. I hope you like it :D

libs/execution/src/lib/types/io-types/filesystem.ts
schlingling: Right, i changed the docs

libs/execution/src/lib/types/io-types/filesystem-node-directory.ts
schlingling: Now it works equally, so no track in new issue neccesary.

schlingling:
> To avoid the `instanceof` checks, consider introducing new methods in `FileSystemNode` that perform the recursive descents. Those can then be called in the `FileSystem` class.
> 
> Also I think that paths should generally be treated case-sensitive.

all implemented with new commits

felix-oq (approved):
A few more hints for improving the implementation, but otherwise this is good to go.

libs/extensions/std/exec/src/archive-interpreter-executor.ts
```diff
@@ -2,6 +2,7 @@
 //
 // SPDX-License-Identifier: AGPL-3.0-only
 
+import assert = require('assert');
```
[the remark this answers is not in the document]
schlingling: changed

libs/execution/src/lib/types/io-types/filesystem-node.ts
```diff
@@ -0,0 +1,10 @@
+// SPDX-FileCopyrightText: 2023 Friedrich-Alexander-Universitat Erlangen-Nurnberg
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+
+export abstract class FileSystemNode {
+  constructor(public name: string) {}
+  abstract addChild(fileSystemNode: FileSystemNode): FileSystemNode | null;
+  abstract getNode(path: string): FileSystemNode | null;
+  abstract putNode(path: string, node: FileSystemNode): FileSystemNode | null;
```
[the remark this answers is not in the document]
schlingling: changed

libs/execution/src/lib/types/io-types/filesystem-node-directory.ts
```diff
@@ -0,0 +1,66 @@
...
+    for (const child of this.children) {
+      const f = child.getNode(rest.join('/'));
+      if (f) {
+        return f;
+      }
+    }
+    return null;
+  }
+
+  override putNode(path: string, node: FileSystemNode): FileSystemNode | null {
```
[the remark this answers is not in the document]
schlingling: splitted a little bit. But would prefer, having all necessary information on one view, since it is a recursive function. Added additionally comments to make is easier to read.

felix-oq (approved):

Issue #217, opened 2023-03-17
[FEATURE] Composite Block for GTFS


rhazn:
## User Story
1. As a {user}
2. I want {to have a composite block for handling GTFS data}
3. So that {I don't have to set up a very complex pipeline every time}

## User Acceptance Criteria
- [x] Basic functionality for composite blocks https://github.com/jvalue/jayvee/pull/411
- [x] model validation hints concerning composite blocks https://github.com/jvalue/jayvee/pull/411
- [x] tests #440
- [x] docs #438
- [x] add a GTFS/CSV composite block into the stdlib #437 
- [x] Multiple pipelines should be possible, moved to #439

## Examples
- https://github.com/jvalue/amse-exercises/blob/main/exercise5.jv

## Definitions of Done
- [x] A PR has been opened and accepted
- [x] All user acceptance criteria are met
- [x] All tests are passing

PR #262, opened 2023-04-12
RFC 0008: Builtin blocktypes

felix-oq:
Preparation for #217 

Rendered markdown: https://github.com/jvalue/jayvee/blob/rfc-0008-builtin-blocktypes/rfc/0008-builtin-blocktypes/README.md


georg-schwarz (commented):
Nice, looks really intuitive to me! Also hyped for the extension of this work towards composite blocks! ;-)

rfc/0008-builtin-blocktypes/README.md
```diff
@@ -0,0 +1,209 @@
...
+    property database oftype text;
+    property table oftype text;
+}
+```
+</details>
+
+## Drawbacks
+
+- Does not include validation semantics
+- Does not include a standard for documenting blocktypes in the code
```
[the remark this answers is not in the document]
felix-oq: True, will shift that to the section below :+1:

rfc/0008-builtin-blocktypes/README.md
```diff
@@ -0,0 +1,209 @@
...
+
+    output oftype TextFile;
+}
+
+builtin blocktype CSVInterpreter {
+    input oftype TextFile;
+
+    property delimiter oftype text: ",";
+    property enclosing oftype text: "";
+    property enclosingEscape oftype text: "";
```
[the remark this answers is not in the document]
georg-schwarz: Currently, we could simplify like that. My guts feeling says that it could hamper future changes, though. I'd keep pipeline and blocks as separate concepts.

However, there might be a relation between a `property` (compile-time parameter) and a `parameter` (runtime parameter).

rfc/0008-builtin-blocktypes/README.md
```diff
@@ -0,0 +1,212 @@
...
+  - Produces an error if the blocktype is unknown
+- `input` / `output` keywords to type the input and output of the blocktype
+  - If omitted, a blocktype has no input or output
+- `property` keyword to define a property with a name and a type
+  - Optionally, a default value can be assigned
+- Adds new keywords for existing property types:
+  - Valuetypes: `regex`, `cellrange` / `row` / `column` / `cell`, `valuetype-assignment`
+  - Typed collections: `collection<type>`
+    - Nested collections are not supported for now
+  - IO types: `File`, `FileSystem`, `TextFile`, `Sheet`, `Table`
```
rhazn, line 45: Not sure how I feel about having valuetypes and io types having totally distinct writing styles (lower case with dash separator vs uppercase and camelcase). I would merge them to one style we use in the language already for property names.

rfc/0008-builtin-blocktypes/README.md
```diff
@@ -0,0 +1,212 @@
...
+
+Currently, all blocktypes only exist implicitly in the language.
+To make their definition more explicit, we should support the declaration of builtin blocktypes (i.e. blocktypes that are built into the language).
+This allows users to look up IO types and properties without having to open the documentation pages.
+
+## Explanation
+
+- `builtin` keyword to indicate that the blocktype is built into the language
+  - Produces an error if the blocktype is unknown
+- `input` / `output` keywords to type the input and output of the blocktype
```
rhazn, line 37: I'd encourage inputs and outputs with an array with strings 

inputs: [
  "default" oftype file
]

That would allow us to be more future proof if (when) we move to multiple in/outputs but might be more complex right now. I am open to both, just leaving it as feedback for your consideration. Classic agile SE would teach us to not do future proofing work and just go with single in/outputs for now.

rfc/0008-builtin-blocktypes/README.md
```diff
@@ -0,0 +1,212 @@
...
+    property lineFrom oftype integer;
+    property lineTo oftype integer;
+
+    output oftype TextFile;
+}
+
+builtin blocktype CSVInterpreter {
+    input oftype TextFile;
+
+    property delimiter oftype text: ",";
```
rhazn, line 106: I don't really like this, I don't think this makes it clear that it is a default parameter. Can we do: property delimiter oftype text default ","?

rfc/0008-builtin-blocktypes/README.md
```diff
@@ -0,0 +1,212 @@
...
+    property password oftype text;
+    property database oftype text;
+    property table oftype text;
+}
+```
+</details>
+
+## Drawbacks
+
+- Adds valuetype keywords for properties, but these types cannot be assigned to table columns
```
rhazn, line 177: Does this mean the IOTypes keywords like Table? I would consider these extra M2 concepts and not value types, then being able to use them in oftype but not for table columns is the same as using blocktypes in oftype for blocks bot nowhere else.

rhazn: Ah reading on, I assume this means types like collection, valuetype-assignment etc. I don't think it's a big problem.

rfc/0008-builtin-blocktypes/README.md
```diff
@@ -0,0 +1,212 @@
...
+- No mixing of property valuetypes and column valuetypes
+- Using a less verbose syntax (e.g. omit `property` keyword or use a shorthand operator instead of `oftype` keyword)
+
+## Possible Future Changes/Enhancements
+
+- Usage of valuetypes for typing properties
+  - Constraints can then be used to validate property values
+  - Properties can be considered inputs of a block, so property values can be provided dynamically by pipes
+- Include validation semantics beyond valuetypes
+- Introduce a standard for documenting blocktypes in the code
```
rhazn, line 191: I really like this, Python has the concept of method docs as comment in the method, Eiffel has the pre and post conditions of methods in separate parts. I think we could really do something good by structuring and requiring documentation for the blocks, their inputs/outputs and examples like we have right now directly in the jayvee code.

rfc/0008-builtin-blocktypes/README.md
```diff
@@ -0,0 +1,212 @@
...
+- Possibility to declare multiple named inputs and outputs for blocktypes
+- Can serve as a foundation for composite blocktypes, e.g.:
+
+```jayvee
+composite blocktype HttpFileExtractor {
```
rhazn, line 196: Love it.

rhazn (approved):

rfc/0008-builtin-blocktypes/README.md
felix-oq: I'd emphasize to avoid introducing more "artificial" collection properties to M2 concepts (like the `constraints` collection in valuetypes) that are not defined in the code.

I'd rather use a syntax similar to properties to define multiple inputs / outputs:
```
// default input:
input oftype sheet;

// additional named input:
input myInput oftype sheet;
```

And finally I want to highlight that modelling multiple inputs/outputs is considered beyond the scope of this RFC.

rfc/0008-builtin-blocktypes/README.md
felix-oq: I think the syntax fits well, because it uses our assignment operator `:` to assign a default value to a property (similar to how values are assigned to properties in blocks). What about generally changing `:` to `=` for assignments, so it is more obvious what happens?

rfc/0008-builtin-blocktypes/README.md
felix-oq: Sure, we can use the same cases :+1:

Then it would be `valuetypeAssignment` instead of `valuetype-assignment`, right?

rhazn: Yes. But also fileSystem for io types.

rfc/0008-builtin-blocktypes/README.md
rhazn: I think it would be weird to have two collection type syntaxes. Even though I can see the difference here with the additional named input. But the same could have been said for constraints for example. I am happy with going with the single input oftype sheet for now in any case though, let's commit to that and not discuss multiple inputs here you are right.

rfc/0008-builtin-blocktypes/README.md
rhazn: I disagree a lot with that, the left hand expression of the colon is very different between a property in a block (where it is just the property name) and here (where it is a longish description of a property that will be in the block. I think the character used is irrelevant for that concept, I just do not see a value assignment here. In a block, the left hand of the assignment solely refers to a property name, here, without the default value assignment the line is a long definition of a property with attributes (the name delimiter and a supertype).

This will also be the only value assignment in the blocktype syntax (input/outputs do not use it).

georg-schwarz: I think Felix' approach is more imperative (property declaration + initialization / value assignment) coming from the classic notion of programming languages where you can initialize object properties already in the class, while Philip's approach is more declarative (declares property and its default value) where the object does the actual initialization incorporating the declared default values.

The imperative approach is closer to other programming languages while the declarative approach is easier for non-programmers. Personally, I'd also lean towards the declarative approach (`default` keyword) but I understand the beauty of the imperative approach as well as it leads to a less bloated language.

In general, we should discuss whether human-readability / suitability for novices is one of our core USPs or not. This should also guide us in design decisions like this one.

rfc/0008-builtin-blocktypes/README.md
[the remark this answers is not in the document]
georg-schwarz: Not sure what I tried to say here... My best guess: `property` and `parameter` should follow a similar syntax (even though they are not the same thing)

rfc/0008-builtin-blocktypes/README.md
rhazn: I can not follow that, my point is that there is a fundamental difference between assigning a value to a property e.g. in our blocks:

```
.... {
    propertyName: value;
}
``` 

And here, where we do not have a block that has propertyNames with value assignments but a definition of a propertytype that, to be more readable, we do inline:

```
.... {
    property delimiter oftype text;
}
``` 

You could write this, following our usual syntax and not inline, as:

```
property {
    name: delimiter;
    supertype: text;
}
```

Then my point is, the default value is a property of the propertytype:

```
property {
    name: delimiter;
    supertype: text;
    defaultValue: ","
}
```

and not a value assignment, therefor the syntax of having a value assigment to the inline propertytype definition makes no sense.

Note that in all of these cases we have not imperatively gave any machine any commands but only described what properties the propertytype has.

georg-schwarz: I think the other viewpoint is the following. Let's assume we implement this in Java, then it would look like this:

```java
class MyNewBlockType extends AbstractBlockType {
    string delimiter = ","; // declare and initialize with default value
    
    // aliternative: declaration and initialization separately
    string delimiter;
    
    MyNewBlockType() {
      delimiter = ",";
    }
}

var newBlock = new MyNewBlockType();
newBlock.delimiter = ";"; // overwrite default value
```

It is close to the object-oriented/imperative programming paradigm instead of the declarative one.
Obviously, if we look into the meta-model, there you would have a class like this or so.

```java
class Attribute<T> {
    string identifier;
    T valuetype; // dunno which type fits here
    T defaultValue;
}
```

So maybe we are just discussing what level we want to present to the Jayvee user?

rhazn: Agreed on assignment syntax in peer debriefing meeting.

georg-schwarz (approved):
LGTM. We should probably look into the docs as well for the io-type case changes, although we don't have a real documentation therefore yet.

felix-oq:
> Btw, I'm missing a mechanism to add documentation (description and examples) that we previously had in the `BlockMetaInf` mechanism. Any ideas how that would be approached?

For now, I think it would be fine to leave that aspect in the langauge server (i.e. the language server "magically" provides documentation for builtin block types).

On the long run, we could use structured comments to describe documentation, similar to how it's done in programming languages. Recently I've noticed, that Langium offers some kind of JSDoc support, but I haven't looked into it so far: https://github.com/langium/langium/blob/main/packages/langium/CHANGELOG.md#jsdoc-support

rhazn:
We should merge this, right @georg-schwarz ? I can take over creating an implementation issue.

rfc/0008-builtin-blocktypes/README.md
```diff
@@ -0,0 +1,212 @@
...
+SPDX-License-Identifier: AGPL-3.0-only
+-->
+
+# RFC 0008: Builtin blocktypes
+
+|             |                      |
+|-------------|----------------------|
+| Feature Tag | `builtin-blocktypes` |
+| Status      | `DISCUSSION`              | <!-- Possible values: DRAFT, DISCUSSION, ACCEPTED, REJECTED -->
+| Responsible | `@felix-oq`          |
```
georg-schwarz, line 13: We should probably change the responsible person.

georg-schwarz (approved):

rfc/0008-builtin-blocktypes/README.md
rhazn: Done in the latest commit, I think then we can go like this, right? If we need to iterate on this I'd rather create a new PR.

PR #270, opened 2023-04-13
Extend cell writer to support multiple cell writes

### LINKED ISSUE #214: [FEATURE] Deal easier with renaming header names
## User Story
1. As a {user}
2. I want {to be able to have an easier time renaming header names in files}
3. So that {I can parse them}

## User Acceptance Criteria
- [ ] a header name mapper/renamer block is available that allows to map multiple input names to output names to mass translate the headers
- [ ] a block is available that renames multiple headers/cells instead of chaining multiple cell writers (to rename the non-unique header values)

## Examples
- This example includes non-unique headers (so some need to be renamed) and german headers (that should be bulk translated) https://github.com/jvalue/amse-exercises/blob/main/exercise4.jv

## Definitions of Done
- [ ] All tests are passing

felix-oq:
Closes #214

### DIFF
--- a/libs/language-server/src/lib/ast/wrappers/cell-range-wrapper.ts
+++ b/libs/language-server/src/lib/ast/wrappers/cell-range-wrapper.ts
@@ -138,6 +138,37 @@ export class CellRangeWrapper<N extends CellRangeLiteral = CellRangeLiteral>
     return cellRange.cellFrom !== undefined && cellRange.cellTo !== undefined;
   }
 
+  isOneDimensional(): boolean {
+    return (
+      this.from.columnIndex === this.to.columnIndex ||
+      this.from.rowIndex === this.to.rowIndex
+    );
+  }
+
+  numberOfCells(): number {
+    const numberOfRows = this.getNumberOfRows();
+    const numberOfColumns = this.getNumberOfColumns();
+
+    return numberOfRows * numberOfColumns;
+  }
+
+  private getNumberOfRows(): number {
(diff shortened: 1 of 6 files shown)


libs/language-server/src/lib/ast/wrappers/cell-range-wrapper.ts
```diff
@@ -138,6 +138,37 @@ export class CellRangeWrapper<N extends CellRangeLiteral = CellRangeLiteral>
...
+
+  numberOfCells(): number {
+    const numberOfRows = this.getNumberOfRows();
+    const numberOfColumns = this.getNumberOfColumns();
+
+    return numberOfRows * numberOfColumns;
+  }
+
+  private getNumberOfRows(): number {
+    if (this.from.rowIndex === LAST_INDEX && this.to.rowIndex === LAST_INDEX) {
```
rhazn, line 156: Do we need this? If this check is true, line 159 returns 1 as well (LAST_INDEX minus itself is 0, plus 1 will be 1). Same for number of columns.

rhazn (approved):
👍

libs/language-server/src/lib/ast/wrappers/cell-range-wrapper.ts
felix-oq: Yes: `LAST_INDEX` holds the value `Number.POSITIVE_INFINITY`, so without the check we'd compute `Number.POSITIVE_INFINITY - Number.POSITIVE_INFINITY` which yields `NaN`.

Issue #213, opened 2023-03-17
[FEATURE] Value transformations


rhazn:
## User Story
1. As a {user}
2. I want {to solve exercise 4}
3. So that {I can get a good grade in AMSE}

## User Acceptance Criteria
- [ ] Students read in temperature as Celsius and write it as Fahrenheit
- [ ] Students can read in booleans from 'Ja'/'Nein' strings and write it as boolean
- [ ] Students can read in timespans from ms and write it as seconds

## Notes
- Consider value type readers and writers for the first two UAC

## Examples
- see data in https://github.com/jvalue/amse-exercises/blob/main/exercise4.jv

## Definitions of Done
- [ ] A PR has been opened and accepted
- [ ] All user acceptance criteria are met
- [ ] All tests are passing

PR #272, opened 2023-04-14
Introduce boolean expressions

felix-oq:
Preparation for #213

Allows users to use boolean expressions wherever a boolean value is expected. Introduces the following operators (ordered descending by [precedence](https://en.wikipedia.org/wiki/Order_of_operations)):

- `not`: unary operator to negate a boolean value
- `==` / `!=`: binary operator to check (in)equality
- `xor`: binary operator to compute exclusive OR
- `and`: binary operator to compute AND
- `or`: binary operator to compute OR

Example for a boolean expression: `(not true or false) xor true and true` which evaluates to `true`.

Also adds a diagnostic to propose a simplification for expressions that can be evaluated at compile-time:
image

### DIFF
--- a/libs/language-server/src/lib/ast/model-util.ts
+++ b/libs/language-server/src/lib/ast/model-util.ts
@@ -11,10 +11,13 @@ import { getMetaInformation } from '../meta-information/meta-inf-registry';
 
 import {
   BlockDefinition,
+  BooleanExpression,
   PipelineDefinition,
   PrimitiveValuetypeKeywordLiteral,
   PropertyValueLiteral,
   ValuetypeDefinitionReference,
+  isBinaryExpression,
+  isBooleanExpression,
   isBooleanLiteral,
   isCellRangeLiteral,
   isCollectionLiteral,
@@ -22,6 +25,7 @@ import {
   isNumericLiteral,
   isRegexLiteral,
   isTextLiteral,
+  isUnaryExpression,
   isValuetypeAssignmentLiteral,
   isValuetypeDefinitionReference,
 } from './generated/ast';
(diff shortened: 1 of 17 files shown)


georg-schwarz (approved):
Awesome!

libs/language-server/src/lib/ast/model-util.ts
```diff
@@ -209,15 +216,68 @@ export function inferTypesFromValue(
...
+        return !evaluateExpression(expression.expression);
+      }
+      default:
+        assertUnreachable(unaryOperator);
+    }
+  }
+  if (isBinaryExpression(expression)) {
+    const binaryOperator = expression.operator;
+    switch (binaryOperator) {
+      case '==': {
```
rhazn, line 243: Imho these should be constants, not strings.

rhazn (approved):

libs/language-server/src/lib/ast/model-util.ts
felix-oq: Note that they are not hardcoded. Instead, they are compliant with the literal type of the generated AST.

Excerpt from the generated `ast.ts`:
``` typescript
export interface BinaryExpression extends AstNode {
    readonly $container: BinaryExpression | CollectionLiteral | PropertyAssignment | UnaryExpression | ValuetypeDefinition;
    readonly $type: 'BinaryExpression';
    left: BooleanExpression
    operator: '!=' | '==' | 'and' | 'or' | 'xor'
    right: BooleanExpression
}
```

rhazn: Not really the only thing enums would be good for but in that case I don't feel bad about doing it like this, makes sense 👍

PR #273, opened 2023-04-17
RFC 0009: Simple Expressions

georg-schwarz:


rfc/0009-expressions/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+### Handling of numbers
+
+Division and remainder operators **always** produce `decimal` values. To convert to an integer value, the operators `floor`, `ceil`, `round` have to be used.
+
+Division by zero should produce an `INVALID` value in the future. For now, it will just throw an error. 
+
+### Interplay of booleans and numbers
+
+We infer the datatype based on the operator. The resulting value type of each operator unambiguous.
```
rhazn, line 72: "is unambiguous"?

rfc/0009-expressions/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+Division and remainder operators **always** produce `decimal` values. To convert to an integer value, the operators `floor`, `ceil`, `round` have to be used.
+
+Division by zero should produce an `INVALID` value in the future. For now, it will just throw an error. 
+
+### Interplay of booleans and numbers
+
+We infer the datatype based on the operator. The resulting value type of each operator unambiguous. 
+
+The operands `==` and `!=` are the only ones that allow operands of different types. Other operators are unambiguous. Comparing the equality of two operands of different value types always resolves to `false`.
```
rhazn, line 74: How about `3/3 == floor (3/3)`? For the record I am fine with:

- Comparing equality of two operands of different value types to always throw an error
- Comparing equality of two operands of different value types to always resolve to `false` and raise a warning

rfc/0009-expressions/README.md
```diff
@@ -0,0 +1,94 @@
...
+- Some operators are functions in other languages which might be unintuitive.
+- Always converting to decimals on division might lead to misleading results if the programmer is not aware.
+
+## Alternatives
+
+- Use an established scripting language instead (e.g. Lua script).
+- Automatically convert to integer values based on some criteria.
+- Make `pow` and `root` prefix operators, e.g., `root(5, 2)`.
+- Require grouping brackets on unary operators.
+- Equality check could automatically convert value types into each other if possible.
```
rhazn, line 89: No.

rfc/0009-expressions/README.md
```diff
@@ -0,0 +1,94 @@
...
+- binary operators (see below)
+
+#### Unary Prefix Operatory
+Binary prefix operators follow the pattern `op a`. Brackets can be used for grouping larger sub-expressions.
+- `floor`, `ceil`, `round` for conversion of decimal numbers to integers.
+- `not` for unary inversion of boolean values
+
+#### Binary Operators
+Binary infix operators follow the infix pattern `a op b`.
+- `pow`, `root` for power and root calculation. E.g., `2 pow 3` is 2³, `5 root 2` is `sqrt(5).
```
rhazn, line 48: Imho we should directly include a `sqrt` unary operator for the very common square roots.

rhazn (approved):

rfc/0009-expressions/README.md
```diff
@@ -0,0 +1,94 @@
...
+- Use an established scripting language instead (e.g. Lua script).
+- Automatically convert to integer values based on some criteria.
+- Make `pow` and `root` prefix operators, e.g., `root(5, 2)`.
+- Require grouping brackets on unary operators.
+- Equality check could automatically convert value types into each other if possible.
+
+## Possible Future Changes/Enhancements
+
+- Incorporate operators for `text` values (append, split, get length, ...).
+- Access input data (e.g., sheet cells) to implement value transformations (#213).
```
felix-oq, line 94: Two more points:

- Boolean expressions could be used to describe custom constraints
- Depending on our future design decisions, value transformations may relate more to value representation concepts (e.g. value readers / writers) rather than blocks.

felix-oq (approved):

rfc/0009-expressions/README.md
georg-schwarz: Makes sense to me. Any objections @rhazn ?

rhazn: Making it work I think is good. Calling it sub-type is not very good if it is not practically also a sub-type (e.g. if our built-in valuetype integer is not a sub-type of decimal). But in general, I am okay with error but allowing comparison of decimals and integers with a warning.

georg-schwarz: Phrased it like this: [0fa04db](https://github.com/jvalue/jayvee/pull/273/commits/0fa04db7edd618099aa046eb10f6649d60306a18)

PR #276, opened 2023-04-18
Introduce validation context

felix-oq:
Introduces a `ValidationContext` class that is provided instead of `ValidationAcceptor` to report diagnostics during validation. More precisely it is a wrapper around the usual `ValidationAcceptor`. The purpose is to allow checking whether errors occurred during validation and use this information to abort the validation early.

The PR also includes a few improvements for existing validation checks.

### DIFF
--- a/libs/language-server/src/lib/validation/checks/block-definition.ts
+++ b/libs/language-server/src/lib/validation/checks/block-definition.ts
@@ -6,7 +6,7 @@
  * See the FAQ section of README.md for an explanation why the following ESLint rule is disabled for this file.
  */
 /* eslint-disable @typescript-eslint/no-unnecessary-condition */
-import { ValidationAcceptor, assertUnreachable } from 'langium';
+import { assertUnreachable } from 'langium';
 
 import {
   BlockDefinition,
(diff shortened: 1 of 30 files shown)


libs/language-server/src/lib/validation/checks/block-definition.ts
```diff
@@ -6,35 +6,43 @@
...
-  checkPipesOfBlock(block, 'input', accept);
-  checkPipesOfBlock(block, 'output', accept);
+  checkBlockType(block, context);
+  if (context.hasErrorOccurred()) {
+    return;
+  }
+
+  checkPipesOfBlock(block, 'input', context);
+  checkPipesOfBlock(block, 'output', context);
 }
```
felix-oq, line 34: Here is an example use case for the new `ValidationContext`.

libs/language-server/src/lib/validation/validation-context.ts
```diff
@@ -0,0 +1,26 @@
+// SPDX-FileCopyrightText: 2023 Friedrich-Alexander-Universitat Erlangen-Nurnberg
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+
+import { AstNode, DiagnosticInfo, ValidationAcceptor } from 'langium';
+
+export class ValidationContext {
+  private errorOccurred = false;
```
georg-schwarz, line 8: We could also count the `errors`, `warnings`, etc. and then have derived attributes/methods for `hasErrorOccured` / `hasWarningOccured`. Not sure if we need such a thing soon, though.

libs/language-server/src/lib/validation/validation-registry.ts
```diff
@@ -18,25 +24,41 @@ import { validatePropertyBody } from './checks/property-body';
...
+) => MaybePromise<void>;
+
+function wrapCheck<T extends AstNode = AstNode>(
+  check: JayveeValidationCheck<T>,
+): ValidationCheck<T> {
+  return (node: T, accept: ValidationAcceptor): MaybePromise<void> => {
+    const context = new ValidationContext(accept);
+    return check(node, context);
+  };
+}
```
georg-schwarz, line 64: This design says that each validation registered is independent from each other, right? Just wanna be sure :)

libs/language-server/src/lib/validation/validation-context.ts
felix-oq: Yeah, also thought about that, but for now a boolean flag is sufficient. Can be added if needed in the future :+1:

georg-schwarz (approved):
Nice work!

libs/language-server/src/lib/validation/validation-registry.ts
felix-oq: Yes, now there is at most one validation function for each kind of AST node. That single validation function can now perform checks as needed and potentially abort.

The `warpCheck` function you highlighted here is used to initialize the `ValidationContext`, so each validation function gets a "fresh" context.

PR #278, opened 2023-04-19
Show constraint docs when hovering

felix-oq:
Now shows a description when hovering over a constraint type:

image


georg-schwarz (approved):
Nice!

PR #279, opened 2023-04-20
Add an empty standard library

felix-oq:
Preparation for #217 

Adds a builtin Jayvee file `standard.jv` that can serve as a standard library in the future. For now, the file is empty (it only contains a comment, so it has no semantics yet).

In VS Code, when navigating to that file, it is marked as read-only and thus cannot be edited by users.

To accomplish this, I followed this guide from the Langium documentation: https://langium.org/guides/builtin-library/

### DIFF
--- a/apps/vs-code-extension/src/standard-library-file-system-provider.ts
+++ b/apps/vs-code-extension/src/standard-library-file-system-provider.ts
@@ -1,0 +1,80 @@
+// SPDX-FileCopyrightText: 2023 Friedrich-Alexander-Universitat Erlangen-Nurnberg
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+
+import { STANDARD_LIBRARY_SOURCECODE } from '@jvalue/jayvee-language-server';
+import {
+  EventEmitter,
+  ExtensionContext,
+  FileChangeEvent,
+  FileStat,
+  FileSystemError,
+  FileSystemProvider,
+  FileType,
+  Uri,
+  workspace,
+} from 'vscode';
+
+export class StandardLibraryFileSystemProvider implements FileSystemProvider {
+  // The following class members only serve to satisfy the interface:
+  private readonly didChangeFile = new EventEmitter<FileChangeEvent[]>();
+  onDidChangeFile = this.didChangeFile.event;
+
+  static register(context: ExtensionContext) {
+    context.subscriptions.push(
+      workspace.registerFileSystemProvider(
+        'builtin',
+        new StandardLibraryFileSystemProvider(),
+        {
+          isReadonly: true,
+          isCaseSensitive: false,
+        },
+      ),
+    );
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  stat(uri: Uri): FileStat {
+    const date = Date.now();
+    return {
+      ctime: date,
+      mtime: date,
+      size: STANDARD_LIBRARY_SOURCECODE.length,
+      type: FileType.File,
+    };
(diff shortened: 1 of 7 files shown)


libs/language-server/src/lib/builtin-library/jayvee-standard-library.ts
```diff
@@ -0,0 +1,9 @@
+// SPDX-FileCopyrightText: 2023 Friedrich-Alexander-Universitat Erlangen-Nurnberg
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+
+export const STANDARD_LIBRARY_FILENAME = 'standard.jv';
+
+export const STANDARD_LIBRARY_SOURCECODE = `
+// TBD
```
georg-schwarz, line 8: Some loose end here?

libs/language-server/src/lib/builtin-library/jayvee-workspace-manager.ts
```diff
@@ -0,0 +1,39 @@
...
+    collector: (document: LangiumDocument) => void,
+  ): Promise<void> {
+    collector(
+      this.documentFactory.fromString(
+        STANDARD_LIBRARY_SOURCECODE,
+        URI.parse(`builtin:///${STANDARD_LIBRARY_FILENAME}`),
+      ),
+    );
+    return Promise.resolve();
+  }
```
georg-schwarz, line 38: Do we need to delegate here to super? Or is it a real override?

apps/vs-code-extension/src/standard-library-file-system-provider.ts
```diff
@@ -0,0 +1,80 @@
...
+      mtime: date,
+      size: STANDARD_LIBRARY_SOURCECODE.length,
+      type: FileType.File,
+    };
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  readFile(uri: Uri): Uint8Array {
+    // We could return different libraries based on the URI
+    // We have only one, so we always return the same
```
georg-schwarz, line 50: Nice comments, love it. Was exactly my question when I read the code.

apps/vs-code-extension/src/standard-library-file-system-provider.ts
```diff
@@ -0,0 +1,80 @@
...
+    );
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  stat(uri: Uri): FileStat {
+    const date = Date.now();
+    return {
+      ctime: date,
+      mtime: date,
+      size: STANDARD_LIBRARY_SOURCECODE.length,
```
georg-schwarz, line 42: Would have expected this is the byte size of the document? Then we'd have to do some calculation like each character times bytes per character.

georg-schwarz (approved):
Quite hard to review for me as a lot of interaction with langium happens. I'm not sure if the API is used correctly, but the code looks good an well documented for me ;-)

libs/language-server/src/lib/builtin-library/jayvee-workspace-manager.ts
felix-oq: The default implementation only returns an empty promise. But probably still safer to call `super` at the beginning :+1:

libs/language-server/src/lib/builtin-library/jayvee-standard-library.ts
felix-oq: No, the std library is supposed to be empty, so it contains just that comment on purpose.

PR #280, opened 2023-04-20
RFC 0010 - Expressions as Constraints

georg-schwarz:
Here is the rendered markdown for better reading experience:
https://github.com/jvalue/jayvee/tree/rfc-0010-expression-constraints/rfc/0010-constraint-expressions

### DIFF
--- a/rfc/0010-constraint-expressions/README.md
+++ b/rfc/0010-constraint-expressions/README.md
@@ -91,21 +91,21 @@ constraint IFOPT_Format on text = value matches /[a-z]{2}:\d+:\d+(:\d+)?(:\d+)?/
 
 ### Hidden Enhancements
 * Expressions allow parameters (keyword `value`).
-* The length of string parameters is accessible by `value.length`.
+* The length of text parameters is accessible by `value.length`. Evaluates to an integer (zero or positive).
 * The assignment operator `=` instead of `:` because the latter looks off in that context.
-* The binary string operator `matches` is introduced.
+* The binary operator `matches` is introduced which evaluates to a boolean value. The usage is `<text> matches <regex>`.
 
 ### Usage in ValueTypes
-In ValueTypes, these constraints can be added like before by name. Combining constraints on non-matching value types leads to an error at runtime in the IDE hints.
+In ValueTypes, these constraints can be added like before by name. Combining constraints on non-matching value types leads to a diagnostic error at compile-time (by the language server).
 ```
-valuetype GasFillLevel oftype number {
+valuetype GasFillLevel oftype decimal {
     constraints: [ GasFillLevelRange ];
 }
 ```
 
-Alternatively, constraints can be defined inline with the caveat that they are not reusable elsewhere. Value then adheres to the type of the ValueType.
+Alternatively, constraints can be defined inline with the caveat that they are not reusable elsewhere. Then the value, that the `value` keyword refers to, adheres to the type of the enclosing ValueType.
 ```
-valuetype GasFillLevel oftype number {
+valuetype GasFillLevel oftype decimal {
     constraints: [ value >= 0 and value <= 100 ];
 }
 ```
--- a/rfc/0010-constraint-expressions/README.md
+++ b/rfc/0010-constraint-expressions/README.md
@@ -42,7 +42,7 @@ Instead of using this block-like syntax, we can use expressions to define constr
 
 The general syntax looks like this: 
 ```
-constraint <name> on <primitive value type> = <expression with parameter value, evaluating to boolean>
+constraint <name> on <primitive value type>: <expression with parameter value, evaluating to boolean>
 ```
 
 ### Examples
@@ -52,7 +52,7 @@ The following subsections will go over the existing constraints and showcase the
 #### AllowlistConstraint
 
 ```
-constraint TimeUnitString on text = 
+constraint TimeUnitString on text:
   value =="ms" or 
   value == "s" or 
   value == "min" or 
@@ -64,11 +64,11 @@ constraint TimeUnitString on text =
 
 #### DenylistConstraint
 ```
-constraint NoPrimaryColors on text = 
+constraint NoPrimaryColors on text:
   not (value == "red" or value == "blue" or value == "yellow");
 
 // alternative
-constraint NoPrimaryColors on text = 
+constraint NoPrimaryColors on text:
   value != "red" and 
   value != "blue" and 
   value != "yellow";
@@ -76,17 +76,20 @@ constraint NoPrimaryColors on text =
 
 #### LengthConstraint
 ```
-constraint JavaStringLength on text = value.length  >= 0 and value.length  <= maxLength;
+constraint JavaStringLength on text:
+  value.length  >= 0 and value.length  <= maxLength;
 ```
 
 #### RangeConstraint
 ```
-constraint HundredScale on decimal = value > 1 and value <= 100; 
+constraint HundredScale on decimal:
+  value > 1 and value <= 100; 
 ```
 
 #### RegexConstraint
 ```
-constraint IFOPT_Format on text = value matches /[a-z]{2}:\d+:\d+(:\d+)?(:\d+)?/;
+constraint IFOPT_Format on text:
+  value matches /[a-z]{2}:\d+:\d+(:\d+)?(:\d+)?/;
 ```
 
 ### Hidden Enhancements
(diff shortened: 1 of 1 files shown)


rfc/0010-constraint-expressions/README.md
```diff
@@ -0,0 +1,132 @@
...
+#### RegexConstraint
+```
+constraint IFOPT_Format on text = value matches /[a-z]{2}:\d+:\d+(:\d+)?(:\d+)?/;
+```
+
+### Hidden Enhancements
+* Expressions allow parameters (keyword `value`).
+* The length of string parameters is accessible by `value.length`.
+* The assignment operator `=` instead of `:` because the latter looks off in that context.
+* The binary string operator `matches` is introduced.
```
felix-oq, line 96:
```suggestion
* The binary operator `matches` is introduced which evaluates to a boolean value. The usage is `<text> matches <regex>`.
```

rfc/0010-constraint-expressions/README.md
```diff
@@ -0,0 +1,132 @@
...
+
+### Usage in ValueTypes
+In ValueTypes, these constraints can be added like before by name. Combining constraints on non-matching value types leads to an error at runtime in the IDE hints.
```
felix-oq, line 99: Reporting such errors at runtime is unnecessarily late, it can be done statically by the language server:
```suggestion
In ValueTypes, these constraints can be added like before by name. Combining constraints on non-matching value types leads to a diagnostic error at compile-time (by the language server).
```

rfc/0010-constraint-expressions/README.md
```diff
@@ -0,0 +1,132 @@
...
+```
+valuetype GasFillLevel oftype number {
+    constraints: [ GasFillLevelRange ];
+}
+```
+
+Alternatively, constraints can be defined inline with the caveat that they are not reusable elsewhere. Value then adheres to the type of the ValueType.
```
felix-oq, line 106:
```suggestion
Alternatively, constraints can be defined inline with the caveat that they are not reusable elsewhere. Then the value, that the `value` keyword refers to, adheres to the type of the enclosing ValueType.
```

rfc/0010-constraint-expressions/README.md
```diff
@@ -0,0 +1,132 @@
...
+```
+valuetype GasFillLevel oftype number {
```
felix-oq, line 108:
```suggestion
valuetype GasFillLevel oftype decimal {
```

rfc/0010-constraint-expressions/README.md
```diff
@@ -0,0 +1,132 @@
...
+### Hidden Enhancements
+* Expressions allow parameters (keyword `value`).
+* The length of string parameters is accessible by `value.length`.
+* The assignment operator `=` instead of `:` because the latter looks off in that context.
+* The binary string operator `matches` is introduced.
+
+### Usage in ValueTypes
+In ValueTypes, these constraints can be added like before by name. Combining constraints on non-matching value types leads to an error at runtime in the IDE hints.
+```
+valuetype GasFillLevel oftype number {
```
felix-oq, line 101:
```suggestion
valuetype GasFillLevel oftype decimal {
```

rfc/0010-constraint-expressions/README.md
```diff
@@ -0,0 +1,132 @@
...
+    constraints: [ value >= 0 and value <= 100 ];
+}
+```
+
+
+## Drawbacks
+
+- Inconsistency of constraint definition via `=` and property initialization via `:`.
+- Future constraints have to be expressible by a boolean expression.
```
felix-oq, line 117: Not necessarily if we also keep the current concept for configurable constraints. It would be well suited for more complex / builtin constraints (as elaborated above).

rfc/0010-constraint-expressions/README.md
```diff
@@ -0,0 +1,132 @@
...
+- Users now have to be able to write expressions instead of just supplying property values.
```
felix-oq, line 118: Will be solved via #273 and is not a drawback in my opinion.

rfc/0010-constraint-expressions/README.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Alternatives
+
+- Syntax to define parameter in constraint could be different, e.g., `constraint MyConstraint = (number) => <expression>`.
+- Allow if/else statements in a more imperative way rather than packing everything into one expression.
```
felix-oq, line 123: I think this could indeed be beneficial for cases where the expression becomes large and confusing. An alternative would be to nest constraints, so the overall complexity could be broken up into smaller sub-constraints.

As this is not an exclusive alternative, I'd shift this point to the `Possible Future Changes/Enhancements` section.

For now, I'd just go with the boolean expressions suggested in this RFC and observe whether the need for splitting up complex expressions arises at some point.

rfc/0010-constraint-expressions/README.md
```diff
@@ -0,0 +1,132 @@
...
+- Put the expression into a `{}`-wrapped scope for consistency (probably together with if/else statements, looks more like a method).
+
+
+
+## Possible Future Changes/Enhancements
+- Allow renaming of `value` to a custom parameter name, e.g., `for text as x`.
```
felix-oq, line 129: Do you have a use case in mind where this would be beneficial?

rfc/0010-constraint-expressions/README.md
```diff
@@ -0,0 +1,132 @@
...
+- add `in` operator for collections for syntactic sugar.
```
felix-oq, line 130: I think this is generally a good idea and it should be incorporated in this RFC. An `in` operator would make it easier to define constraints similar to the current `AllowlistConstraint` and `DenylistConstraint`:
```jayvee
constraint NoPrimaryColors on text =
  not (value in ["red", "blue", "yellow"]);
```

rfc/0010-constraint-expressions/README.md
```diff
@@ -0,0 +1,132 @@
...
+- Allow users to define own `ConstraintTypes` that allow the current syntax again.
```
felix-oq, line 131: In case we decide to also keep our current concept, this would be a nice way to incorporate the two approaches (boolean expressions on the one hand and configurable constraint types on the other).

felix-oq (approved):
All in all I'm pretty happy with this RFC. It gives users way more flexibility on how they can define constraints without just relying on a limited number of builtin constraint types. Also, static analysis can be used to suggest simplifications for such expressions, resulting in simpler and more readable code.

rfc/0010-constraint-expressions/README.md
```diff
@@ -0,0 +1,132 @@
...
+
+#### RegexConstraint
+```
+constraint IFOPT_Format on text = value matches /[a-z]{2}:\d+:\d+(:\d+)?(:\d+)?/;
+```
+
+### Hidden Enhancements
+* Expressions allow parameters (keyword `value`).
+* The length of string parameters is accessible by `value.length`.
+* The assignment operator `=` instead of `:` because the latter looks off in that context.
```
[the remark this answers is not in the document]
georg-schwarz: Yes, probably not a change to do in disguise in this PR.

rfc/0010-constraint-expressions/README.md
georg-schwarz: Sure. I think it can be beneficial for the code's readability as custom parameter names are more expressive while  `value` stay generic. Second use-case related to composite value types where - depending on how we will implement that - having multiple named parts could be referable in constraints. Not problematic if we require to prepend the `value.` for dereferencing, though.

rfc/0010-constraint-expressions/README.md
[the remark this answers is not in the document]
rhazn: I dislike using `=` here, I'd go with `:` in the context of this RFC. If we want to switch to `=` we should do so in a separate decision I think.

rfc/0010-constraint-expressions/README.md
rhazn: I'd go with value and dereferencing for now but I can follow your reasoning.

rfc/0010-constraint-expressions/README.md
```diff
@@ -0,0 +1,132 @@
...
+```
+
+#### LengthConstraint
+```
+constraint JavaStringLength on text = value.length  >= 0 and value.length  <= maxLength;
+```
+
+#### RangeConstraint
+```
+constraint HundredScale on decimal = value > 1 and value <= 100;
```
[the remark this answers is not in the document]
rhazn: I agree on not doing it here but I want to mention that I consider that syntax a big plus for intuitive use. A value between two limits comes up surprisingly often and is weird to express in classic boolean checks.

rhazn (approved):
Looks great 👍

rfc/0010-constraint-expressions/README.md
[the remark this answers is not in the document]
georg-schwarz: => https://github.com/jvalue/jayvee/issues/288

rfc/0010-constraint-expressions/README.md
georg-schwarz: => https://github.com/jvalue/jayvee/issues/289

PR #283, opened 2023-04-21
RFC 0011 - Value Transformers

georg-schwarz:
Rendered markdown [here](https://github.com/jvalue/jayvee/tree/rfc-0011-value-transformer/rfc/0011-value-transformer)

### DIFF
--- a/rfc/0011-value-transformer/README.md
+++ b/rfc/0011-value-transformer/README.md
@@ -9,7 +9,7 @@ SPDX-License-Identifier: AGPL-3.0-only
 | | |
 |---|---|
 | Feature Tag | `value-transform` |
-| Status | `DISCUSSION` | <!-- Possible values: DRAFT, DISCUSSION, ACCEPTED, REJECTED -->
+| Status | `ACCEPTED` | <!-- Possible values: DRAFT, DISCUSSION, ACCEPTED, REJECTED -->
 | Responsible | `georg-schwarz` |
 <!-- 
   Status Overview:


rfc/0011-value-transformer/README.md
```diff
@@ -0,0 +1,191 @@
...
+**Note:** This RFC focuses enhancing the current state but has an elaborate "Possible Enhancements" section that showcases how this concept can be extended in the future.
+
+## Motivation
+
+Right now, we can only "parse" values but not manipulate them in order to clean data.
+
+## Explanation
+
+```
+transformer <name> {
```
rhazn, line 35: The syntax is very close to a block, I don't think that is a good idea.

Can we consider `from`/`to` for input and output?

Also, is a transformer just the singular M2 concept? We might want to implement custom transformers on a language level, I'd consider making this one type of transformer, an `ExpressionTransformer` for example. Combined, this would become:

```
transformer <name> oftype ExpressionTransformer {
  from <inputName> oftype <valuetype>; // potentially many
  to <outputName> oftype <valuetype>; // for now: exactly one

  <outputName>: <expression producing output valuetype>;
}

File: rfc/0011-value-transformer/README.md [hunk 2]
```diff
@@ -0,0 +1,191 @@
...
+### Usage 1: TableInterpreter
+
+The `TableInterpreter` block uses a transformer from `text` to the desired value type with the `with` keyword. Parsing can still be used as currently (`as` keyword)
+
+Example:
+```
+block GasReserveTableInterpreter oftype TableInterpreter {
+  header: true;
+  columns: [ // only allow transformers with one input and one output or plain parsing here
+      "Datum" with DateTransformer,
```
rhazn, line 95: I don't feel it is a good idea to bring transformers into this, we should focus on only using valuetypes consistently.

So I think, this line should always be:

`"Datum" as date`

For defining a transformer, I think it could be okay to apply it here with `with` or `using`:

`"Datum" as date using DateTransformer`

If possible, we could try to find a chain of transformers that automatically work for the input type to the output type. 

The core of my objection here is not using valuetypes consistently though.

rfc/0011-value-transformer/README.md
```diff
@@ -0,0 +1,191 @@
...
+```
+block GermanToBooleanBlock oftype TableTransformerBlock {
+  inputColumns: ["Kritisch"];  // array to satisfy all inputs of transformer
+  outputColumns: ["IsCritical"]; // might also overwrite one of the input columns
+  use: GermanToBooleanTransformer;
+}
+```
+
+### Usage 3: SheetTransformerBlock
+There is a newly introduced `SheetTransformerBlock` block type that applies the mapper to a set of cells in a `sheet`. Please note that the value type is not stored in sheets and thus lost after computation. Outputs are therefore parsed to string automatically by the block.
```
rhazn, line 121: I think this shows a difference in the semantics. One is changing valuetypes into other valuetypes, one is changing the structure of a string. If the valuetype is lost after the transformation, I do not think we should use transformators for it, it should be a different concept that can only map from string to string (?).

rfc/0011-value-transformer/README.md
rhazn: In general, I thing we could replace valuetype statements with valuetypes and transformers? Consider:

- A valuetype A
- A valuetype B
- A transformer from A to B

`<valuetypeA>` is currently a valid valuetype expression. We could replace it with `<valuetypeB> using <transformerAtoB>`. That would lead to exactly my suggested syntax and could be used in any place we currently use valuetypes. It also automatically allows chaining: `<valuetypeC> using <transformerBtoC> using <transformerAtoB>`.

rfc/0011-value-transformer/README.md
```diff
@@ -0,0 +1,191 @@
...
+Storing interim results can improve coding experience.
+```
+transformer <name> {
+  // inputs and outputs
+  var <varBane> oftype <valuetype>: <expression producing valuetype>; // alternative: infer type automatically 
+  // use variable to compute outputs
+}
+```
+
+### Composite ValueTypes
```
rhazn, line 159: I think this overall makes sense, the suggestion kind of naturally extends to composite valuetypes.

rfc/0011-value-transformer/README.md
```diff
@@ -0,0 +1,191 @@
...
+  input houseNumber oftype text;
+  output address oftype Address; // has properties "streetName" oftype StreetName and "houseNumber" oftype HouseNumber 
+
+  // "." notation to write on property of composite value type
+  address.streetName: streetName using StreetNameReader; // StreetNameReader is other transformation text -> StreetName
+  address.houseNumber: houseNumber using HouseNumberReader; // HouseNumberReader is other transformation text -> HouseNumber
+}
+```
+
+### Mappers
```
rhazn, line 175: This has to be called `match on` I beg you.

rfc/0011-value-transformer/README.md
```diff
@@ -0,0 +1,191 @@
+<!--
+SPDX-FileCopyrightText: 2023 Friedrich-Alexander-Universitat Erlangen-Nurnberg
+
+SPDX-License-Identifier: AGPL-3.0-only
+-->
+
+# RFC 0000: Value Transformer
```
rhazn, line 7: 0011

rfc/0011-value-transformer/README.md
```diff
@@ -0,0 +1,191 @@
...
+
+| | |
+|---|---|
+| Feature Tag | `value-transformer` |
+| Status | `DRAFT` | <!-- Possible values: DRAFT, DISCUSSION, ACCEPTED, REJECTED -->
```
rhazn, line 12: DISCUSSION

rhazn (approved):
Sorry this took so long. I am very unhappy with the TableInterpreter example, the rest of the comments I'd consider optional.

rfc/0011-value-transformer/README.md
georg-schwarz: Good point. I could imagine sth like this as well.

rfc/0011-value-transformer/README.md
georg-schwarz: Let me rephrase to check if I understood correctly:

We introduce a concept `ValueTypeExpression` that always evaluates to a `ValueType`. The syntax is as follows:
```
<final output valuetype> using <transformer from input to output type> (using <transformer from input to output type>)*
```
We would introduce a set of default transormers that allow parsing from text to the primitive ValueType.

rfc/0011-value-transformer/README.md
georg-schwarz: Or allow only writing to string? Then the user would need to chain a default transformer to convert back to string.

rfc/0011-value-transformer/README.md
georg-schwarz: [789fba5](https://github.com/jvalue/jayvee/pull/283/commits/789fba58202bbf7fef9967816122a4c5b932804f)

rfc/0011-value-transformer/README.md
georg-schwarz: [789fba5](https://github.com/jvalue/jayvee/pull/283/commits/789fba58202bbf7fef9967816122a4c5b932804f)

rfc/0011-value-transformer/README.md
felix-oq: Regarding composite blocks, that may be added in the future, the keywords `input` and `output` could fit there as well. In case their semantics will be the same (or similar), I'd go with `input` and `output` for both concepts. Otherwise I think `from` and `to` are fine too.

---

Not sure about the transformer type part. Would the `mapper` concept drafted below be considered a potential other type besides `ExpressionTransformer`?

Should a user be able to create own transformer types? If not, a distinction via a keyword (e.g. `transformer` and `mapper`) would be suited better in my opinion.

rfc/0011-value-transformer/README.md
```diff
@@ -0,0 +1,191 @@
...
+
+### Usage 3: SheetTransformerBlock
+There is a newly introduced `SheetTransformerBlock` block type that applies the mapper to a set of cells in a `sheet`. Please note that the value type is not stored in sheets and thus lost after computation. Outputs are therefore parsed to string automatically by the block.
+
+Example:
+```
+block GermanToBooleanBlock oftype SheetTransformerBlock {
+  inputCells: [range A3:A17]; // if multiple arguments: have to match in dimensions
+  outputCells: [range D3:D17]; // has to match dimension of inputs
+  use: GermanToBooleanTransformer;
```
felix-oq, line 128: I assume, the transformer here is also required to have one input and one output? And the same transformation is applied to every cell?

Or are the input cells supposed to be the distinct inputs of the given transformer (e.g. 3 input cells for a transformer with 3 inputs)? In that case, how would 2-dimensional cell ranges be treated?

rfc/0011-value-transformer/README.md
```diff
@@ -0,0 +1,191 @@
...
+
+
+## Drawbacks
+- declaration of in and outputs might be much boilerplate.
+- TableInterpreter blocks only allow transformers with one input and one output.
+
+## Alternatives
+
+- use other keyword, e.g., `vtrans`, `mapper`, ...
+- shorten `input` and `output` keywords to `in` and `out`
```
felix-oq, line 140: Definitely something to consider if we get respective feedback.

felix-oq (approved):

rfc/0011-value-transformer/README.md
rhazn: I'd start with using and move to nicer syntax later.

And I agree on the `ValueTypeExpression`. Unsure what you mean by default transformers, if you refer to built-in valuetypes I'd just have them auto parse as now. Primitive valuetypes are user defined so I am unsure how we would introduce default transformers for them?

rfc/0011-value-transformer/README.md
rhazn: Didn't even remember `from`/`to` is technically like pipes :D. I still like it more because it feels more intuitive to me for transforming something.

Re: Types, I would probably consider mappers other types of transformers. But I can see how users would not be able to add new transformer types so it would be okay to leave out the oftype (?). Just going with the transformer keyword has my blessing at least.

georg-schwarz:
I reduced the scope of this RFC to only cover one input and one output. Everything should go into separate incremental RFCs. 

Any changes from your sides before I merge? @rhazn @felix-oq

rhazn:
No, looks good 👍

felix-oq:
Also fine for me :+1:

PR #284, opened 2023-04-23
BUGFIX comments and improvements

### LINKED ISSUE #300: [BUG] GTFS static/RT example errors
## Steps to reproduce
1. Execute `/example/gtfs-static-and-rt.jv` with jayvee interpreter v 0.0.14

## Description
- Expected: Works
- Actual:
```
[GTFSExtractor] Fetching raw data from https://ratpdev-mosaic-prod-bucket-raw.s3-eu-west-1.amazonaws.com/11/exports/1/gtfs.zip
[GTFSExtractor] Successfully fetched raw data
[GTFSExtractor] Execution duration: 531 ms.
[ZipArchiveInterpreter] Loading zip file from binary content
[ZipArchiveInterpreter] Execution duration: 78 ms.
error: Unexpected Error The expression evaluated to a falsy value:

  (0, assert_1.strict)(addedFile != null)
 occured during processing
In <filename>:11:8
11 |     block ZipArchiveInterpreter oftype ArchiveInterpreter {
```
## Notes
https://github.com/jvalue/jayvee/blob/ae26874313028c9473808af318d20a8c89811603/libs/extensions/std/exec/src/archive-interpreter-executor.ts#L87

schlingling:
Fixed a bug explained in the comment, added some more comments and renamed gtfs-related pipeline.

closes #300


libs/extensions/std/exec/src/archive-interpreter-executor.ts
```diff
@@ -83,7 +83,7 @@ export class ArchiveInterpreterExecutor
             mimeType,
             content,
           );
-          const addedFile = root.putFile(relPath, file);
+          const addedFile = root.putFile('/' + relPath, file);
```
schlingling, line 86: We need to at a root-indicator here, since we defined in `InMemoryFileSystem`, that all paths should start at root. JSZip return here the relative Path starting from the zipped folder, so adding the root-indicator manually should be fine. --> Now the gtfs pipelines work again.

libs/execution/src/lib/types/io-types/filesystem-node-directory.ts
```diff
@@ -12,12 +12,12 @@ export class FileSystemDirectory extends FileSystemNode {
 
   override getNode(pathParts: string[]): FileSystemNode | null {
     const [firstPart, ...rest] = pathParts;
-    // Base case: We called a wrong node
+    // Base case: Called a wrong noce
```
rhazn, line 15:
```suggestion
    // Base case: Called a wrong node
```

libs/extensions/std/exec/src/archive-interpreter-executor.ts
rhazn: This probably also should use a constant like `InMemoryFileSystem.PATH_SEPARATOR`?

rhazn (approved):

libs/extensions/std/exec/src/archive-interpreter-executor.ts
schlingling: Yeah, to support this, i have to add an getter in the inmemoryfilesystem.

libs/execution/src/lib/types/io-types/filesystem-inmemory.ts
```diff
@@ -51,8 +51,12 @@ export class InMemoryFileSystem implements FileSystem {
     return null;
   }
 
+  getPathSeparator(): string {
+    return InMemoryFileSystem.PATH_SEPARATOR;
+  }
+
   private processPath(path: string): string[] | null {
-    if (!path.startsWith('/')) {
+    if (!path.startsWith(InMemoryFileSystem.PATH_SEPARATOR)) {
```
rhazn, line 59: If you add a getter function you must use it consistently. You can make `PATH_SEPARATOR` private and `getPathSeparator` `static`. And then use the getter function everywhere.

rhazn (approved):

schlingling:
closes #300

libs/execution/src/lib/types/io-types/filesystem-inmemory.ts
schlingling: made getter static and now in consistent use.

rhazn (approved):

PR #290, opened 2023-04-25
Implement expressions

felix-oq:
Preparation for #213, implementation of [RFC 0009](https://github.com/jvalue/jayvee/pull/273)

Introduces new operators, type inference and evaluation semantics for expressions. Also includes options for `lazy` and `exhaustive` evaluation (affects binary boolean operators so they don't necessarily evaluate both sub-expressions).

In many cases, the code for type inference is pretty much the same per operator, except for some special cases (e.g. different treatment of `decimal` and `integer` and that operands of `==`/`!=` need to have the same type). I'm not sure how to properly refactor the code, so both usual and special cases can be covered conveniently. Introducing a [sub-type relationship](https://en.wikipedia.org/wiki/Subtyping) between `integer` and `decimal` or an implicit conversion mechanism could already help to generally cope with the first special case. Let me know if you have further ideas, maybe we can discuss that in a call.

The prior redundancy in the evaluation was mostly eliminated in https://github.com/jvalue/jayvee/pull/290/commits/02f05bddc0025896f7aaf9ad9dd679842e963370 by introducing evaluator classes that implement shared evaluation logic. A similar refactoring for type inference also address the problem described above.

Known issue:

Sometimes, when using symbolic operators without surrounding spaces, a lexer error occurs:
image
image

Not sure why that is the case (probably due to the grammar?). I fiddled around with it a bit but didn't find an immediate solution for the problem.


libs/language-server/src/lib/ast/expressions/operators/binary-arithmetic-expression.ts
```diff
@@ -0,0 +1,202 @@
...
+      }
+      if (!isNumericType(rightType)) {
+        context?.accept(
+          'error',
+          generateUnexpectedTypeMessage(numericTypes, rightType),
+          {
+            node: expression.right,
+          },
+        );
+      }
```
georg-schwarz, line 56: Could be extracted to a separate method to improve the reading flow of the method. Personally, I think nested control structures are somewhat hard to read.

libs/language-server/src/lib/ast/expressions/operators/binary-arithmetic-expression.ts
```diff
@@ -0,0 +1,202 @@
...
+> = (
+  expression: BinaryExpression,
+  strategy: EvaluationStrategy,
+  context: ValidationContext | undefined,
+): boolean | number | string | undefined => {
+  assert(expression.operator === '*');
+  const leftValue = evaluateExpression(expression.left, strategy, context);
+  if (leftValue === undefined && strategy === EvaluationStrategy.LAZY) {
+    return undefined;
+  }
```
georg-schwarz, line 80: Ah I see, lazy evaluation introduces the return type `undefined`.

libs/language-server/src/lib/validation/checks/property-assignment.ts
```diff
@@ -0,0 +1,130 @@
...
+      context.accept(
+        'error',
+        `Runtime parameters are not allowed for properties of type ${propertyType}`,
+        {
+          node: property,
+          property: 'name',
+        },
+      );
+    }
+  } else {
```
georg-schwarz, line 82: Use a `return` statement in the case above to get rid of this else (?)

libs/language-server/src/lib/validation/checks/property-assignment.ts
```diff
@@ -0,0 +1,130 @@
...
+          `The value needs to be of type ${propertyType} but is of type ${matchingPropertyType}`,
+          {
+            node: property,
+            property: 'value',
+          },
+        );
+      } else {
+        if (isExpression(propertyValue)) {
+          checkExpressionSimplification(propertyValue, context);
+        }
```
georg-schwarz, line 103: This is the "happy" path, so it should be on the function's indent level I think.

Generally, it feels easier to read when everything on the "happy path" is on the function's indent level and every "unhappy" path deviates by being further indented. But only my personal preference :)

libs/language-server/src/lib/ast/expressions/operators/binary-arithmetic-expression.ts
felix-oq: The return type `undefined` is not caused by lazy evaluation but rather by arithmetic errors (e.g. division by zero). This code construct skips the evaluation of the right operand if the left one evaluated to `undefined`.

And sure, I can move the evaluation strategy part to the front. I agree that it's probably easier to read that way.

libs/language-server/src/lib/ast/expressions/operator-evaluator.ts
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    assert(this.leftValueTypeguard(leftValue));
+    assert(this.rightValueTypeguard(rightValue));
+
+    return this.doEvaluate(leftValue, rightValue, expression, context);
+  }
+}
+
+export abstract class BooleanShortCircuitOperatorEvaluator
```
georg-schwarz, line 131: A comment here to explain what "short circuit" means might be helpful :)

libs/language-server/src/lib/ast/expressions/operator-evaluator.ts
```diff
@@ -0,0 +1,169 @@
...
+    private readonly leftValueTypeguard: OperandValueTypeguard<L>,
+    private readonly rightValueTypeguard: OperandValueTypeguard<R>,
+  ) {}
+
+  protected abstract doEvaluate(
+    leftValue: L,
+    rightValue: R,
+    expression: BinaryExpression,
+    context: ValidationContext | undefined,
+  ): T | undefined;
```
georg-schwarz, line 107: If I understand it correctly, we return `undefined` on failure. (a) We should document that, and (b) we should have it on the radar to change it as soon as we have expressions that can evaluate to undefined (if that ever should be the case).

georg-schwarz (approved):
Good job on the refactoring!

From a code review perspective it looks good to me, but can't guarantee that I was able to catch all bugs (esp. because the featiure is quite large). I feel like we should start introducing tests soon with the increasing logic.

libs/execution/src/lib/constraints/executors/allowlist-constraint-executor.ts
```diff
@@ -16,8 +20,12 @@ export class AllowlistConstraintExecutor implements ConstraintExecutor {
       return false;
     }
 
-    const allowlist = context.getTextCollectionPropertyValue('allowlist');
-    const allowlistValues = allowlist.map((e) => e.value);
+    const allowlist = context.getExpressionCollectionPropertyValue('allowlist');
+    const allowlistValues = allowlist.map((expression) => {
```
rhazn, line 24: Imho could be written nicer if you split up the individual steps more:
 
```
  allowlist
    .map(evaluateExpression)
    .filter(value => typeof value === 'string')
    .includes(value)
```

Not needed of course but just personal preference.

libs/extensions/std/lang/src/text-file-interpreter-meta-inf.ts
```diff
@@ -25,18 +26,16 @@ export class TextFileInterpreterMetaInformation extends BlockMetaInformation {
           },
           validation: (property, context) => {
             const propertyValue = property.value;
-            assert(isTextLiteral(propertyValue));
+            assert(isExpression(propertyValue));
```
rhazn, line 29: This seems to be repeated quite a bit, might be good to have a util method to get property values of a specific type. I.e. `evaluatedPropertyExpression<T>` (in this case `evaluatedPropertyExpression<string>` that reads the property, evaluates the expression, asserts the type and returns the correctly typed result.

libs/language-server/src/lib/ast/model-util.ts
```diff
@@ -189,93 +189,47 @@ export function runtimeParameterAllowedForType(
...
+  PropertyValuetype.DECIMAL,
+];
+export function isNumericType(
+  type: PropertyValuetype | undefined,
+): type is PropertyValuetype.INTEGER | PropertyValuetype.DECIMAL {
+  if (type === undefined) {
+    return false;
+  }
+  return (
+    type === PropertyValuetype.INTEGER || type === PropertyValuetype.DECIMAL
```
rhazn, line 206: How does this not use the previously defined numericTypes const?

rhazn (approved):
Sick 👍

libs/execution/src/lib/constraints/executors/allowlist-constraint-executor.ts
felix-oq: I'd like to keep the `assert` runtime check rather than silently filtering non-string values (which should never happen at this point). So I'll keep the current implementation.

libs/extensions/std/lang/src/text-file-interpreter-meta-inf.ts
felix-oq: https://github.com/jvalue/jayvee/pull/290/commits/4c63e1dbfd4d90495c741b04c3dffcd9662d79ff

PR #296, opened 2023-04-28
Use root version number for projects

felix-oq:
Changes the setup so that the version from the root `package.json` is used for all published projects.

Also includes some fixes for broken scripts that I noticed along the way.


apps/interpreter/package.json
```diff
@@ -18,5 +17,6 @@
   "publishConfig": {
     "access": "public",
     "registry": "https://registry.npmjs.org"
-  }
+  },
+  "version": "will be overridden by the root version"
```
georg-schwarz, line 21: Interesting idea for documentation :)
You sure it works?

georg-schwarz: Is inconsistent with other apps where the field was just removed.

tools/scripts/add-package-json-version.mjs
```diff
@@ -0,0 +1,20 @@
...
+const [, , projectName] = process.argv;
+
+process.chdir(workspaceRoot);
+const rootPackageJson = parsePackageJson();
+
+process.chdir(getOutputPath(projectName));
+
+const packageJson = parsePackageJson();
+packageJson.version = rootPackageJson.version;
+writePackageJson(packageJson);
```
georg-schwarz, line 20: My understanding: write to the `package.json` in the output directory, so not into the source one. Would we need to write into the package.lock there as well? Or does it not exist in the output dir?

apps/interpreter/package.json
felix-oq: Maybe something like `the version field here is required for the interpreter, but the value will be overridden by the root version` would be better?

georg-schwarz: Aaah! Makes sense, thx! :)

tools/scripts/add-package-json-version.mjs
felix-oq: You're right with your understanding. `package-lock.json` files don't exist in the output dirs.

apps/interpreter/package.json
felix-oq: You're welcome! Would it be better that way?

georg-schwarz: Would turn it around, but yes :)

georg-schwarz (approved):

PR #305, opened 2023-04-30
RFC 0013 Lossless Arithmetic

rhazn:
This RFC is rather short on actual suggestions which I plan to extend after feedback, maybe someone already knows a good JS library to handle this? But I would like to establish a decision on our goals, be mathematically correct in expressions or accept the interpreter limitations.


rfc/0013-lossless-arithmetic/0013-lossless-arithmetic.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Motivation
+
+Jayvee is a language to model data pipelines and will therefore handle numeric data as well. For this data, it is important that it arrives in a sink as correctly as possible. Number representation and evaluation of arithmetric expressions has various edge cases in any programming languages, especially in Javascript. Our goal should be to be as correct as possible.
+
+## Explanation
+
+### Use appropriate data types for numbers
+
+[BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) should be used for `integer`s instead of numbers.
```
georg-schwarz, line 35: Found this library that might be a good entry point to this topic besides the native `BigInt`: https://github.com/MikeMcl/decimal.js/

georg-schwarz:
I totally agree. We should clearly formulate the limitations of each primitive value type.

rhazn:
I would accept this and create a task to gather more info on how to implement this, could you update your review @georg-schwarz ?

rfc/0013-lossless-arithmetic/0013-lossless-arithmetic.md
```diff
@@ -0,0 +1,49 @@
...
```
georg-schwarz, line 35: Rather than describing the implementation of the solution, we should describe the consequences for the language IMO. For example, failing validations for numbers we cannot parse correctly, e.g.,  `3.00000000000000000000000000000001` since it would be parsed as a plain `3`.

georg-schwarz (changes requested):
While I generally agree with the idea of this RFC, I think we should phrase it differently in the explanation section (see below).

georg-schwarz (approved):

PR #307, opened 2023-05-02
Introduce type computers and implicit conversion

felix-oq:
Follow up for #290 

Refactors the type inference by introducing `TypeComputer` classes. They share an abstract base class for common operator behavior (i.e. expecting operands of specific types and computing the resulting type from them). There is also an interface for more flexibility. It is needed by the equality operator, because it has exceptional semantics (takes operands of any types but they need to be the same type).

Also adds a simple mechanism for implicit conversion (in opposition to what was agreed upon in #273). This simplifies the implementation but removes the warnings when having operands of different numeric types `integer` and `decimal`.


libs/language-server/src/lib/ast/expressions/operator-type-computer.ts
```diff
@@ -0,0 +1,131 @@
+// SPDX-FileCopyrightText: 2023 Friedrich-Alexander-Universitat Erlangen-Nurnberg
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+
+import { ValidationContext } from '../../validation/validation-context';
+import { BinaryExpression, UnaryExpression } from '../generated/ast';
+// eslint-disable-next-line import/no-cycle
+import { PropertyValuetype } from '../model-util';
+
+export interface UnaryOperatorTypeComputer {
```
georg-schwarz, line 10: Adding a short documenting JSdoc to the interface methods might make sense to improve onboarding.

georg-schwarz (approved):
Nice refactor!

PR #310, opened 2023-05-04
Architecture overview docs

felix-oq:
So far just an initial version regarding this topic. Let me know whether something is missing or unclear.


apps/docs/docs/dev/04-architecture-overview.md
```diff
@@ -0,0 +1,42 @@
+---
+title: Architecture overview
+---
+
+Jayvee has a clear separation between the language itself and its execution.
+This guides the overall architecture as you will notice in the next sections.
```
rhazn, line 6: I am not a fan of directly talking to the reader as if you are writing a script for a lecture. I notice we talk directly to the reader in our other docs as well (which I dislike but we should keep for consistency), but we should at least not directly "talk" to the reader. I'd rewrite this like ", this guide gives an overview of the overall architecture.".

Similar for the other texts (e.g., in the language server "there is the" could be replaced by just "the", makes it sound less conversational).

rhazn (approved):
Looks great overall 👍

georg-schwarz (approved):

PR #312, opened 2023-05-05
Move Valuetype Wrapper to the language-server

georg-schwarz:
This PR moves the AstNodeWrapper `Valuetype` to the language-server. 

This refactor will allow building up an *is-subtype* or *is-convertable-to* functionality in the language-server as part of the language instead of part of the execution. Such a similar functionality is currently implemented in the `convertsImplicitlyTo` method.

Changes along the way:
- `isValid` functionality was extracted to a visitor that stays in the execution
- `getStandardRepresentation` functionality was extracted to a separate concept that stays in the execution. Unfortunately, I couldn't solve it as visitor due to the generic return type.


libs/language-server/src/lib/ast/wrappers/value-type/visitors/visitable-valuetype.ts
felix-oq, line 1: I think this interface can be abolished by moving its method to the `Valuetype` interface

libs/execution/src/lib/types/valuetypes/standard-representation.ts
```diff
@@ -0,0 +1,103 @@
...
+  BooleanValuetype,
+  DecimalValuetype,
+  IntegerValuetype,
+  PrimitiveType,
+  TextValuetype,
+  Valuetype,
+} from '@jvalue/jayvee-language-server';
+
+export class StandardRepresentationResolver {
+  private readonly DECIMAL_COMMA_SEPARATOR_REGEX = /^[+-]?([0-9]*[,])?[0-9]+$/;
```
felix-oq, line 16: The same regex is also used in `IsValidVisitor`, so consider extracting it as a global constant.

felix-oq (approved):
LGTM, only left a few minor remarks

libs/execution/src/lib/types/valuetypes/standard-representation.ts
georg-schwarz: Addressed in [c6a0803](https://github.com/jvalue/jayvee/pull/312/commits/c6a08038f19426453a5ad90e66229b861ac2095c)

PR #320, opened 2023-05-10
Implement concept for value transformations [RFC 0011]

### LINKED ISSUE #213: [FEATURE] Value transformations

felix-oq:
Closes #213.

### DIFF
--- a/libs/execution/src/lib/transforms/transform-executor.ts
+++ b/libs/execution/src/lib/transforms/transform-executor.ts
@@ -25,27 +25,22 @@ export class TransformExecutor {
     port: TransformPortDefinition;
     valuetype: Valuetype;
   } {
-    const inputPorts = this.transform.body.ports.filter(
-      (x) => x.kind === 'from',
-    );
-    assert(inputPorts.length === 1);
-    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-    const inputPort = inputPorts[0]!;
-    const inputType = inputPort.valueType;
-    const inputValuetype = createValuetype(inputType);
-
-    return {
-      port: inputPort,
-      valuetype: inputValuetype,
-    };
+    return this.getPortDetails('from');
   }
 
   getOutputDetails(): {
     port: TransformPortDefinition;
     valuetype: Valuetype;
   } {
+    return this.getPortDetails('to');
+  }
+
+  private getPortDetails(kind: TransformPortDefinition['kind']): {
+    port: TransformPortDefinition;
+    valuetype: Valuetype;
+  } {
     const outputPorts = this.transform.body.ports.filter(
-      (x) => x.kind === 'to',
+      (x) => x.kind === kind,
     );
     assert(outputPorts.length === 1);
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
--- a/libs/execution/src/lib/transforms/transform-executor.ts
+++ b/libs/execution/src/lib/transforms/transform-executor.ts
@@ -1,3 +1,7 @@
+// SPDX-FileCopyrightText: 2023 Friedrich-Alexander-Universitat Erlangen-Nurnberg
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+
 import { strict as assert } from 'assert';
 
 import {
--- a/libs/execution/src/lib/transforms/transform-executor.ts
+++ b/libs/execution/src/lib/transforms/transform-executor.ts
@@ -1,0 +1,140 @@
+import { strict as assert } from 'assert';
+
+import {
+  InternalValueRepresentation,
+  TransformDefinition,
+  TransformOutputAssignment,
+  TransformPortDefinition,
+  Valuetype,
+  createValuetype,
+  evaluateExpression,
+} from '@jvalue/jayvee-language-server';
+
+import { ExecutionContext } from '../execution-context';
+import { TableColumn } from '../types/io-types/table';
+import { IsValidVisitor } from '../types/valuetypes/visitors/is-valid-visitor';
+
+export class TransformExecutor {
+  constructor(private readonly transform: TransformDefinition) {}
+
+  getInputDetails(): {
+    port: TransformPortDefinition;
+    valuetype: Valuetype;
+  } {
+    const inputPorts = this.transform.body.ports.filter(
+      (x) => x.kind === 'from',
+    );
+    assert(inputPorts.length === 1);
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const inputPort = inputPorts[0]!;
+    const inputType = inputPort.valueType;
+    const inputValuetype = createValuetype(inputType);
+
+    return {
+      port: inputPort,
+      valuetype: inputValuetype,
+    };
+  }
+
+  getOutputDetails(): {
+    port: TransformPortDefinition;
+    valuetype: Valuetype;
+  } {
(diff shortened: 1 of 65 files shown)


libs/language-server/src/lib/ast/wrappers/value-type/valuetype.ts
```diff
@@ -78,10 +93,18 @@ export abstract class AbstractValuetype implements Valuetype {
     return this.supertype;
   }
 
+  equals(target: Valuetype): boolean {
+    return this.isConvertibleTo(target) && target.isConvertibleTo(this); // TODO: solve with registry and === comparison instead
```
georg-schwarz, line 97: We should replace this by a registry mechanism that enforces singletons and compare with `===`

libs/language-server/src/lib/validation/checks/transform-output-assigment.ts
```diff
@@ -0,0 +1,148 @@
...
+  expression: Expression,
+  context: ValidationContext,
+): void {
+  if (isExpressionLiteral(expression)) {
+    return;
+  }
+
+  const evaluatedExpression = evaluateExpression(
+    expression,
+    new EvaluationContext(), // TODO
```
georg-schwarz, line 113: We should check for simplification on sub-expressions as they might not contain free variables.

libs/execution/src/lib/types/valuetypes/visitors/is-valid-visitor.ts
```diff
@@ -39,7 +41,8 @@ export class IsValidVisitor extends ValuetypeVisitor<boolean> {
       );
     }
 
-    return !Number.isNaN(this.value);
+    const result = !Number.isNaN(this.value);
+    return result;
```
felix-oq, line 45: Can be simplified again:
```suggestion
    return !Number.isNaN(this.value);
```

libs/extensions/tabular/lang/src/lib/table-transformer-meta-inf.ts
```diff
@@ -0,0 +1,30 @@
...
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+
+import {
+  BlockMetaInformation,
+  IOType,
+  PrimitiveValuetypes,
+} from '@jvalue/jayvee-language-server';
+
+export class TableTransformerMetaInformation extends BlockMetaInformation {
```
felix-oq, line 11: The documentation for this block type should be added before merging.

libs/execution/src/lib/types/io-types/table.ts
```diff
@@ -2,33 +2,115 @@
...
+  ): void {
+    const column = this.columns.get(columnName);
+    if (column === undefined) {
+      return;
+    }
+
+    column.values.forEach((value, rowIndex) => {
+      callbackfn(value, rowIndex);
+    });
   }
```
felix-oq, line 114: Can be removed because it's unused.

libs/execution/src/lib/types/io-types/table.ts
```diff
@@ -2,33 +2,115 @@
...
+  }
+
+  addRow(row: TableRow): void {
+    const rowLength = Object.keys(row).length;
+    assert(
+      rowLength === this.columns.size,
+      `Added row has the wrong dimension (expected: ${this.columns.size}, actual: ${rowLength})`,
+    );
+    assert(
+      Object.keys(row).every((x) => this.columns.has(x)),
```
felix-oq, line 50:
```suggestion
      Object.keys(row).every((x) => this.hasColumn(x)),
```

libs/extensions/tabular/exec/src/lib/table-interpreter-executor.ts
```diff
@@ -165,14 +166,16 @@ export class TableInterpreterExecutor
         invalidRow = true;
         return;
       }
-      tableRow[tableColumnIndex] = value;
+      tableRow[columnEntry.columnName] = new StandardRepresentationResolver(
+        value,
+      ).fromValuetype(columnEntry.valuetype);
```
felix-oq, line 171: I'd recommend to split this into smaller steps, so it's more readable.

libs/execution/src/lib/transforms/transform-executor.ts
```diff
@@ -0,0 +1,144 @@
...
+} from '@jvalue/jayvee-language-server';
+
+import { ExecutionContext } from '../execution-context';
+import { TableColumn } from '../types/io-types/table';
+import { IsValidVisitor } from '../types/valuetypes/visitors/is-valid-visitor';
+
+export class TransformExecutor {
+  constructor(private readonly transform: TransformDefinition) {}
+
+  getInputDetails(): {
```
felix-oq, line 24: Could be refactored to avoid code duplication, it's almost similar to `getOutputDetails`.

libs/language-server/src/lib/ast/model-util.ts
```diff
@@ -158,6 +164,44 @@ export function getValuetypeName(
   return valuetype.keyword;
 }
 
+export function inferBasePropertyValuetype(
```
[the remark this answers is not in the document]
georg-schwarz: I think this is the differentiation between different kinds of primitive valuetypes? Some are used for data values at runtime and some are not. We could depict that in the valuetype hierarchy instead.

libs/language-server/src/lib/validation/checks/transform-body.ts
```diff
@@ -0,0 +1,166 @@
...
+function checkSingleInput(
+  transformBody: TransformBody,
+  context: ValidationContext,
+): void {
+  const inputs = transformBody.ports?.filter((x) => x.kind === 'from');
+  if (inputs === undefined) {
+    return undefined;
+  }
+
+  if (inputs.length > 1) {
```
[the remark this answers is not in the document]
georg-schwarz: Yes. Having `transforms` without an input might make sense, though. Probably we should adapt the implementation then?

libs/language-server/src/lib/ast/model-util.ts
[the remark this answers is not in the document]
felix-oq: I noticed that there was no such differentiation. So I decided to remove this function and generally use the `createValuetype` function instead.

libs/language-server/src/lib/validation/checks/transform-body.ts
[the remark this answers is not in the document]
georg-schwarz: Ah well, that's true... Probably, we have to add a check for that as well then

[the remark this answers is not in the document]
georg-schwarz: Jointly decided to got with demanding exactly one input and output for now.

georg-schwarz:
Follow-up task: write page in user docs for the introduced `transform` concept.

georg-schwarz:
Follow-up task: allow multiple input columns for transforms

PR #344, opened 2023-05-24
Language-server tests framework

f3l1x98:
# Description
This is the first draft of tests for the language-server. It contains:
- Test utils for parsing, validating and extracting certain parts out of `LangiumDocument`
- Tests for `validation-utils.ts`
- Tests for `block-definition.ts`
- Tests for `column-id.ts`
- Tests for `pipe-definition.ts`
- Tests for `pipeline-definition.ts`
- Tests for `property-body.ts`
- Tests for examples from the `examples` directory in the root

# Notes for discussions
- **This is a draft so any feedback is welcome and some parts are not final!**
- I had to update the `tsconfig.spec.json` and `tsconfig.lib.json` because otherwise the `test/utils.ts` was not accessible in tests but was compiled during build which crashed building due to missing test dependencies
- `property-body.spec.ts` contains some TODOs due to uncertainties from my side.
- The tests do not contain any specific extensions tests because these will be put into the corresponding extensions
- I might extract some parts (mainly the test utils) into a separate root project if they are necessary in other parts down the line (interpreter, extensions, ...)
- I am investigating whether it is possible to use the examples defined in the extensions for documentation as input for the tests (suggestion of @georg-schwarz )
- I have discussed shortly with @georg-schwarz whether to create a separate test extension in order to be able to create special blocks just for testing -> **The current status is that this is too much effort and too little positive reward**
- Concerning the license compliance check: The `test/utils.ts` contains parts taken from the langium repository so I am unsure about the license there


libs/language-server/src/test/validation.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+      path.resolve(__dirname, '../../../../example/gtfs-static.jv'),
+      'utf-8',
+    );
+
+    // Validate input
+    const validationResult = await validate(text);
+    const diagnostics = validationResult.diagnostics;
+    // Expect 0 errors
+    expect(diagnostics).toHaveLength(0);
+  });
```
georg-schwarz, line 93: These test cases seem to be very similar. Can you extract the path as a variable and one method that does the testing logic via `it.each`?

libs/language-server/src/test/validation/checks/pipe-definition.spec.ts
```diff
@@ -0,0 +1,216 @@
...
+        }
+
+        block CarsLoader oftype SQLiteLoader {
+          table: "Cars";
+          file: "./cars.sqlite";
+        }
+        
+        CarsExtractor -> CarsLoader;
+      }
+      `;
```
georg-schwarz, line 196: I wonder if we should consider pulling the models out as variables at the bottom of the file.
If we can give these variables good names, the tests might be more compact and easier to read.

libs/language-server/src/test/validation/checks/pipe-definition.spec.ts
```diff
@@ -0,0 +1,216 @@
...
+
+      const parseResult = await parse(text);
+
+      const pipe: PipeDefinition = extractPipe(parseResult);
+
+      validatePipeDefinition(
+        pipe,
+        new ValidationContext(validationAcceptorMock),
+      );
```
georg-schwarz, line 177: Here we have the same logic in every method (except setting up the model and doing the assertions). Speaking in the classic 3A pattern (arrange, act, assert), we could pull out the whole "act" as a method making the tests more compact and easier to read if we name the pulled out method well.

georg-schwarz (commented):
Looks already really good! Love it to see that testing will harden Jayvee.

One thing I'm unsure about yet is the structure you chose (separate `src/test` directory). I get your point that otherwise there might be a lot of files flying around, e.g. in the `validation/checks` folder. I'd be interested in @felix-oq's opiniton in that regard.

libs/language-server/src/test/validation/checks/pipe-definition.spec.ts
felix-oq: Or alternatively pull them out as separate `.jv` files and have them as test assets. This may ease migrating their syntax to new Jayvee versions once there is some kind of such migration mechanism in place.

felix-oq (commented):
In the Langium Codebase, I found this test file: https://github.com/langium/langium/blob/main/packages/langium/test/grammar/langium-grammar-validator.test.ts

Each test case in there has a lean structure (arrange, act, assert) and a few of them use a `AstNodeLocator` to pick specific AST nodes by using a path for descending (e.g. `'rules@0/definition/elements@1'`).

Did you consider to write our tests in a similar manner?

libs/language-server/src/test/validation.spec.ts
f3l1x98: Didn't know that this exists, will do 👍

libs/language-server/src/test/validation/checks/block-definition.spec.ts
```diff
@@ -0,0 +1,210 @@
+// SPDX-FileCopyrightText: 2023 Friedrich-Alexander-Universitat Erlangen-Nurnberg
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+
+// eslint-disable-next-line @nrwl/nx/enforce-module-boundaries
+import { StdLangExtension } from '@jvalue/jayvee-extensions/std/lang';
```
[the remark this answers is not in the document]
f3l1x98: @georg-schwarz and I discussed whether to write a custom mock extension, but we did not consider this, maybe we should really write one.
But that would cause additional maintenance for the new extension

f3l1x98:
> In the Langium Codebase, I found this test file: https://github.com/langium/langium/blob/main/packages/langium/test/grammar/langium-grammar-validator.test.ts
> 
> Each test case in there has a lean structure (arrange, act, assert) and a few of them use a `AstNodeLocator` to pick specific AST nodes by using a path for descending (e.g. `'rules@0/definition/elements@1'`).
> 
> Did you consider to write our tests in a similar manner?

I will definitely take a look at the `AstNodeLocator` 👍 

Concerning the 3A structure: I might be mistaken, but aren't my tests already in this format? Except perhaps a few too many newlines.

georg-schwarz:
> Concerning the 3A structure: I might be mistaken, but aren't my tests already in this format? Except perhaps a few too many newlines.

Yes, you already followed that pattern. My comment refers to pulling out the repeating parts (which matches the act part bu chance)

libs/language-server/src/test/utils.ts
```diff
@@ -0,0 +1,154 @@
...
+  PipeDefinition,
+  PipelineDefinition,
+  PropertyBody,
+} from '../lib';
+
+export interface ParseHelperOptions extends BuildOptions {
+  documentUri?: string;
+}
+
+export function parseHelper<T extends AstNode = AstNode>(
```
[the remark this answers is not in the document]
f3l1x98: Split up this file into `langium-utils.ts` (copied code incl. license) and `utils.ts` (self-written utils with own license)

libs/language-server/src/test/validation/checks/pipe-definition.spec.ts
f3l1x98: They have been extracted into own `.jv` files inside `test/assets/*` and are loaded using an utility function in `utils.ts`

libs/language-server/src/test/validation.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+// eslint-disable-next-line @nrwl/nx/enforce-module-boundaries
+import { StdLangExtension } from '@jvalue/jayvee-extensions/std/lang';
+import { AstNode } from 'langium';
+import { NodeFileSystem } from 'langium/node';
+
+import { JayveeServices, createJayveeServices, useExtension } from '../lib';
+
+import { ValidationResult, validationHelper } from './utils';
+
+describe('jv example tests', () => {
```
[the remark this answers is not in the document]
f3l1x98: I have moved them into `@jvalue/jayvee-extensions/std/lang` because in the interpreter there will be no validation tests (and the not yet created integration test will also have no direct validation test)

libs/language-server/src/test/validation/checks/block-definition.spec.ts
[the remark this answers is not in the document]
f3l1x98: I have created a test extension in language-server and removed all usages of StdExtension from language server.
The test extension blocks are also defined with tests in mind so they would not make any sense outside of a test environment.

libs/language-server/src/test/utils.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export const validationAcceptorMockImpl = <N extends AstNode>(
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  severity: 'error' | 'warning' | 'info' | 'hint',
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  message: string,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  info: DiagnosticInfo<N>,
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+) => {};
```
felix-oq, line 18: Can be simplified:
```suggestion
import { AstNode, LangiumDocument, ValidationAcceptor } from 'langium';

// eslint-disable-next-line @typescript-eslint/no-empty-function
export const validationAcceptorMockImpl: ValidationAcceptor = () => {};
```

libs/language-server/src/test/langium-utils.ts
```diff
@@ -0,0 +1,61 @@
+/** ****************************************************************************
+ * Copyright 2021 TypeFox GmbH
+ * This program and the accompanying materials are made available under the
+ * terms of the MIT License, which is available in the project root.
+ ******************************************************************************/
```
felix-oq, line 5: To keep the structure of the REUSE header consistent:
```suggestion
// SPDX-FileCopyrightText: 2021 TypeFox GmbH
//
// SPDX-License-Identifier: MIT

```

Then, we need to add the MIT license to our `LICENSES` folder in the root of the repository. This can be done automatically by running `reuse download MIT` (requires the [REUSE tool](https://github.com/fsfe/reuse-tool) to be installed).

libs/extensions/std/lang/src/example-validation.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+    useExtension(new StdLangExtension());
+    // Create language services
+    services = createJayveeServices(NodeFileSystem).Jayvee;
+    // Create validation helper for language services
+    validate = validationHelper(services);
+  });
+
+  it.each([
+    'cars.jv',
+    'gas.jv',
```
felix-oq, line 38: Note that the `gas.jv` file was replaced by `electric-vehicles.jv` in the main branch.

libs/language-server/src/test/assets/block-definition/invalid-block-as-multiple-pipe-inputs.jv
```diff
@@ -0,0 +1,21 @@
...
+  pipe {
+    from: TestPipeFrom;
+    to: TestPipeTo;
+  }
+
+  pipe {
+    from: TestPipeFrom;
+    to: TestPipeTo;
+  }
+}
```
felix-oq, line 21: I'd recommend to overthink the naming. I'd go with something like this:
```suggestion
pipeline Pipeline {
  block BlockTo oftype TestTableLoader {
  }

  block BlockFrom oftype TestFileExtractor {
  }

  pipe {
    from: BlockFrom;
    to: BlockTo;
  }

  pipe {
    from: BlockFrom;
    to: BlockTo;
  }
}
```

For me, calling blocks `Pipe` seems pretty confusing. Regarding the names of the test block types, it would also be helpful to stick to the established conventions (i.e. `Extractor` for block types with no input and `Loader` for block types with no output).

To make the test more interesting, consider using the arrow pipe syntax for one of the pipes (`A -> B`).

libs/language-server/src/test/extension/lib/test-property-meta-inf.ts
```diff
@@ -0,0 +1,104 @@
...
+              STRING_TYPEGUARD,
+            );
+
+            try {
+              new TextDecoder(encodingValue);
+            } catch (error) {
+              context.accept('error', `Unknown encoding "${encodingValue}"`, {
+                node: propertyValue,
+              });
+            }
```
felix-oq, line 56: The validation logic could be simplified, e.g.
```suggestion
            const value = evaluatePropertyValueExpression(
              propertyValue,
              new EvaluationContext(), // we don't know values of runtime parameters or variables at this point
              STRING_TYPEGUARD,
            );

            if (value !== 'valid') {
              context.accept('error', `Invalid value "${value}"`, {
                node: propertyValue,
              });
            }
```

felix-oq (commented):
A few more suggestions, but otherwise this is already looking pretty good :+1:

libs/extensions/std/lang/src/example-validation.spec.ts
f3l1x98: Yeah I will have to merge soon, but thanks for the heads up 👍

libs/language-server/src/test/assets/block-definition/invalid-block-as-multiple-pipe-inputs.jv
f3l1x98: I have separate tests for chained vs single pipes so I did not deem that necessary.
However I did decide to switch all of these (where possible) to chained pipes to further compact files.

libs/extensions/std/lang/tsconfig.spec.json
```diff
@@ -9,6 +9,7 @@
     "jest.config.ts",
     "src/**/*.test.ts",
     "src/**/*.spec.ts",
-    "src/**/*.d.ts"
+    "src/**/*.d.ts",
+    "src/extension.ts"
```
georg-schwarz, line 13: Do we need this?

f3l1x98: Is removed

PR #345, opened 2023-05-24
Implement expression constraints

### LINKED ISSUE #309: [FEATURE] Expression Constraints (RFC 0010)
## User Story
1. As a {Jayvee user}
2. I want {to use expressions to define constraints}
3. So that {I don't have to write in the verbose constraint syntax}

## User Acceptance Criteria
- [ ] The RFC is fulfilled
- [ ] Documentation is updated

## Notes
- [RFC 0010](https://github.com/jvalue/jayvee/tree/main/rfc/0010-constraint-expressions)
- [PR](https://github.com/jvalue/jayvee/pull/280) of the RFC with all discussions 

## Definitions of Done
- [ ] All tests are passing

felix-oq:
Closes #309

Additionally includes a refactoring regarding `StandardRepresentationResolver` and `IsValidVisitor`:
- `StandardRepresentationResolver` is replaced by the function `parseValueToInternalRepresentation`
  - Tries to converts a `string` into an `InternalValueRepresentation` value according to a given valuetype
  - Uses an underlying visitor called `InternalRepresentationParserVisitor`
- `IsValidVisitor` is replaced by the function `isValidValueRepresentation`
  - Verifies, that a given `InternalValueRepresentation` value matches a given valuetype
    - For `AtomicValuetype` also checks its constraints
  - Uses an underlying visitor called `ValueRepresentationValidityVisitor`

Also, when converting values from a `sheet` into values for a `table`, the processing order is changed:
- _Before_: Check if the `string` value is valid, then parse the `string` value to `InternalValueRepresentation`
- _After_: Try to parse the `string` value to `InternalValueRepresentation`, then check whether it is valid
- Makes sure that the value already is in `InternalValueRepresentation` when checking constraints on it


libs/language-server/src/lib/ast/expressions/evaluation.ts
```diff
@@ -56,18 +62,29 @@ export type InternalValueRepresentation =
...
+    private readonly runtimeParameterValues: Map<
+      string,
+      InternalValueRepresentation
+    > = new Map(),
+    private readonly variableValues: Map<
       string,
       InternalValueRepresentation
     > = new Map(),
-    public variableValues: Map<string, InternalValueRepresentation> = new Map(),
+    private valueKeywordValue:
```
georg-schwarz, line 73: I assume that we have this distinguished handling due to the `length` property, right? Otherwise we could just add it to the `varuableValues` map with the key `value`.

libs/language-server/src/lib/ast/model-util.ts
```diff
@@ -149,3 +149,23 @@ export type BinaryExpressionOperator = BinaryExpression['operator'];
 export type AstTypeGuard<T extends AstNode = AstNode> = (
   obj: unknown,
 ) => obj is T;
+
+/**
+ * In the container hierarchy of the given AST node, searches for the closest AST node that satisfies the given type guard.
```
georg-schwarz, line 154:
```suggestion
 * Recursively goes upwards through the AST until it finds an AST node that satisfies the given type guard.
```

georg-schwarz (approved):
Very impressive! I think the code got a little leaner by your refactorings. Left some comments to understand the implications of the `value` variable design. Can be merged from my side.

libs/language-server/src/lib/ast/expressions/evaluation.ts
felix-oq: Yes indeed :+1:

PR #348, opened 2023-05-30
Enable runtime parameters to be validated

### LINKED ISSUE #332: [BUG] Properties with runtime parameters are not validated
## Steps to reproduce

1. Create a valid Jayvee model
2. Add the following constraint to the model:

```jayvee
constraint Constraint oftype RangeConstraint {
	lowerBound: requires A;
	upperBound: requires B;
}
```

3. Run the model with the command `jv -d -e A=5 -e B=0 ./model.jv`

## Description
- Expected: The error message `The lower bound needs to be smaller or equal to the upper bound` is printed by the interpreter
- Actual: The interpreter reports no error and executes the model

felix-oq:
Closes #332

- I followed [this Langium guide](https://langium.org/docs/configuration-services/#adding-new-services) to create a custom service `RuntimeParameterProvider`
  - It is responsible for holding and providing runtime parameter values
  - I decided that the values in there are be stored in a raw format (i.e. as strings) and that they are parsed into an internal value representation on demand
    - This simplifies the validation of runtime parameter values because their values don't have to be parsed to the correct representation before the actual Jayvee model is parsed
- The language server leaves the `RuntimeParameterProvider` empty and provides no parser for values
- The interpreter sets up the `RuntimeParameterProvider` by adding the runtime parameter values entered in the CLI and by providing the `parseValueToInternalRepresentation` function for parsing
  - Additionally, the interpreter adds validation checks for `RuntimeParameterLiteral` which check the presence and paring of the values provided via the CLI
- Besides `ValidationContext` objects, `EvaluationContext` objects are now passed to validation functions
  - `EvaluationContext` objects are initialized with the `RuntimeParameterProvider` instance, so they are able to provide runtime parameter values
  - The validation functions use the provided `EvaluationContext` object when evaluating property values, so runtime parameters can be evaluated properly


apps/interpreter/src/validation-checks/runtime-parameter-literal.ts
```diff
@@ -0,0 +1,104 @@
...
+}
+
+function checkRuntimeParameterValueParsing(
+  runtimeParameter: RuntimeParameterLiteral,
+  validationContext: ValidationContext,
+  evaluationContext: EvaluationContext,
+) {
+  const containerType: BlockTypeLiteral | ConstraintTypeLiteral | undefined =
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+    runtimeParameter.$container.$container.$container?.type;
```
georg-schwarz, line 63: This looks like a magic line of code for me... Can we add a comment or so? Same for the one below. Or extract to method.

georg-schwarz (approved):

PR #359, opened 2023-06-14
Validation Tests

f3l1x98:
This adds the missing validation tests for
- RangeLiteral
- RegexLiteral
- JayveeModel
- PropertyAssignment
- TransformBody
- TransformOutputAssignment
- TypedConstraintDefinition
- ValuetypeDefinition
- ExpressionConstraintDefinition


felix-oq (commented):
As a first comment, consider using lower-case letters for the input / output ports of transforms. That way the test models follow the naming convention established by the `electric-vehicles.jv` example.

libs/language-server/src/lib/validation/checks/expression-constraint-definition.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+    // Parse function for Jayvee (without validation)
+    parse = parseHelper(services);
+  });
+
+  afterEach(() => {
+    // Reset mock
+    validationAcceptorMock.mockReset();
+  });
+
+  it('should have no error on valid expresion constraint', async () => {
```
felix-oq, line 74:
```suggestion
  it('should have no error on valid expression constraint', async () => {
```

libs/language-server/src/lib/validation/checks/jayvee-model.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+    const text = readJvTestAsset(
+      'jayvee-model/invalid-non-unique-pipelines.jv',
+    );
+
+    await parseAndValidateJayveeModel(text);
+
+    expect(validationAcceptorMock).toHaveBeenCalledTimes(2);
+    expect(validationAcceptorMock).toHaveBeenCalledWith(
+      'error',
+      `The pipelinedefinition name "Pipeline" needs to be unique.`,
```
felix-oq, line 74: I think you actually found a bug here. It message is supposed to state `The pipeline name "Pipeline" needs to be unique`.

Same for transforms, valuetypes and constraints.

libs/language-server/src/lib/validation/checks/property-assignment.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+
+  it('info on non simplifiable property expression', async () => {
+    const text = readJvTestAsset(
+      'property-assignment/valid-uneccessarysimplify-info.jv',
+    );
+
+    await parseAndValidatePropertyAssignment(text);
+
+    expect(validationAcceptorMock).toHaveBeenCalledTimes(0);
+  });
```
felix-oq, line 164: Could you add an additional test case to cover the recently added feature #352?

libs/language-server/src/lib/validation/checks/valuetype-definition.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+  });
+
+  it('error on supertype cycle', async () => {
+    const text = readJvTestAsset(
+      'valuetype-definition/invalid-supertype-cycle.jv',
+    );
+
+    await parseAndValidateValuetypeDefinition(text);
+
+    // TODO FIND OUT HOW THAT WORKS
```
felix-oq: I guess you already found out how to do it, right? Otherwise, feel free to ask for help.

felix-oq (approved):
Awesome, well done!

felix-oq:
> I have one question:
> When exactly is the linking done?
>
> ...
>
> The reason for that is, that during the type-inference of the collection values `expression?.value?.ref` (more precise `ref`) in `inferTypeFromReferenceLiteral` is undefined, which from my understanding can either mean that linking has not been done, or the linking failed.

The linking is done during the creation of the `LangiumDocument`, see [here](https://langium.org/docs/document-lifecycle/) for more details. The reason, it does not link in your case is that valuetypes are not referencable in expressions due to the grammar (except for valuetype assignments). You can have a look at it in [expression.langium](https://github.com/jvalue/jayvee/blob/9c908c28b42061992f1689b622b393d691caab44/libs/language-server/src/grammar/expression.langium#L84-L90). So the linker does not even attempt perform the linking in this case because it is considered illegal.

libs/language-server/src/lib/validation/checks/valuetype-definition.spec.ts
f3l1x98: Yeah found out, forgot to remove TODO :sweat_smile:

f3l1x98:
> > I have one question:
> > When exactly is the linking done?
> > ...
> > The reason for that is, that during the type-inference of the collection values `expression?.value?.ref` (more precise `ref`) in `inferTypeFromReferenceLiteral` is undefined, which from my understanding can either mean that linking has not been done, or the linking failed.
> 
> The linking is done during the creation of the `LangiumDocument`, see [here](https://langium.org/docs/document-lifecycle/) for more details. The reason, it does not link in your case is that valuetypes are not referencable in expressions due to the grammar (except for valuetype assignments). You can have a look at it in [`expression.langium`](https://github.com/jvalue/jayvee/blob/9c908c28b42061992f1689b622b393d691caab44/libs/language-server/src/grammar/expression.langium#L84-L90). So the linker does not even attempt perform the linking in this case because it is considered illegal.

I understand, forgot to check whether the grammar considers this valid.
Thank you :+1:

libs/language-server/src/lib/validation/checks/expression-constraint-definition.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    expect(validationAcceptorMock).toHaveBeenCalledTimes(1);
+    expect(validationAcceptorMock).toHaveBeenCalledWith(
+      'error',
+      `The value needs to be of type boolean but is of type integer`,
+      expect.any(Object),
+    );
+  });
+
+  it('info on simplifiable expression constraint', async () => {
```
georg-schwarz, line 99:
```suggestion
  it('should diagnose info on simplifiable expression constraint', async () => {
```

libs/language-server/src/lib/validation/checks/expression-constraint-definition.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+    const text = readJvTestAsset(
+      'expression-constraint-definition/valid-text-constraint.jv',
+    );
+
+    await parseAndValidateExpressionConstraintDefinition(text);
+
+    expect(validationAcceptorMock).toHaveBeenCalledTimes(0);
+  });
+
+  it('error on incompatible type', async () => {
```
georg-schwarz, line 84:
```suggestion
  it('should diagnose error on incompatible type', async () => {
```

libs/language-server/src/lib/validation/checks/expression-constraint-definition.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+  expectNoParserAndLexerErrors,
+  parseHelper,
+  readJvTestAssetHelper,
+  validationAcceptorMockImpl,
+} from '../../../test';
+import { TestLangExtension } from '../../../test/extension';
+
+import { validateExpressionConstraintDefinition } from './expression-constraint-definition';
+
+describe('expression-constraint-definition validation tests', () => {
```
georg-schwarz, line 27:
```suggestion
describe('Validation of Constraint Definition (expression syntax)', () => {
```

georg-schwarz (commented):
Please make sure to follow the jest naming pattern for consistency. I added one example.

`describe` should always point to the subject under test (the class or broader functionality). It is a test file, so no need to mention it is a test.

`it` should always start with a "should" defining the expected behavior, potentially followed by a "when" later to distinguish different preconditions.

georg-schwarz (commented):
Love it, well done!

libs/language-server/src/lib/validation/checks/property-assignment.spec.ts
felix-oq: That is desired behavior for now. So far, the language server is only capable of suggesting simplifications, where sub-expressions can be evaluated statically (i.e. when expressions don't contain free variables such as `input`). At the moment, the simplification engine has no assumptions on individual operators (e.g. it does not know that `floor(floor(x)) == x`).

libs/language-server/src/lib/validation/checks/jayvee-model.spec.ts
georg-schwarz: Do you want to directly fix the bug? :)

georg-schwarz (approved):

libs/language-server/src/lib/validation/checks/jayvee-model.spec.ts
f3l1x98: @georg-schwarz not sure if you mean me, but if yes I would do that in a separate PR and create an issue for that :smile:

PR #363, opened 2023-06-19
In Operator

georg-schwarz:
Introduces an `in` operator that checks for element inclusion in an array.

### DIFF
--- a/libs/language-server/src/lib/ast/expressions/type-computers/in-operator-type-computer.ts
+++ b/libs/language-server/src/lib/ast/expressions/type-computers/in-operator-type-computer.ts
@@ -62,7 +62,16 @@ export class InOperatorTypeComputer implements BinaryOperatorTypeComputer {
         isCollectionValuetype(rightOperandType, PrimitiveValuetypes.Text),
     );
 
-    if (!leftOperandType.isConvertibleTo(rightOperandType.elementType)) {
+    // allow 3 in [3.5, 5.3]
+    const isLeftConvertibleToRight = leftOperandType.isConvertibleTo(
+      rightOperandType.elementType,
+    );
+
+    // allow 3.5 in [3.0, 2.0]
+    const isRightConvertibleToLeft =
+      rightOperandType.elementType.isConvertibleTo(leftOperandType);
+
+    if (!isLeftConvertibleToRight && !isRightConvertibleToLeft) {
       context?.accept(
         'error',
         `The type of the left operand needs to be compatible to the collection element type of the right operand but they differ (left: ${leftOperandType.getName()}, right: ${rightOperandType.elementType.getName()})`,
--- a/libs/language-server/src/lib/ast/expressions/type-computers/in-operator-type-computer.ts
+++ b/libs/language-server/src/lib/ast/expressions/type-computers/in-operator-type-computer.ts
@@ -2,22 +2,27 @@
 //
 // SPDX-License-Identifier: AGPL-3.0-only
 
+import { strict as assert } from 'assert';
+
 import { ValidationContext } from '../../../validation/validation-context';
 import { BinaryExpression } from '../../generated/ast';
-import { type Valuetype } from '../../wrappers/value-type';
 // eslint-disable-next-line import/no-cycle
+import {
+  CollectionValuetype,
+  type Valuetype,
+  isCollectionValuetype,
+} from '../../wrappers/value-type';
 import { PrimitiveValuetypes } from '../../wrappers/value-type/primitive/primitive-valuetypes';
 import { BinaryOperatorTypeComputer } from '../operator-type-computer';
 
-export class EqualityOperatorTypeComputer
-  implements BinaryOperatorTypeComputer
-{
-  private readonly ALLOWED_OPERAND_TYPES: Valuetype[] = [
-    PrimitiveValuetypes.Boolean,
+export class InOperatorTypeComputer implements BinaryOperatorTypeComputer {
+  private readonly ALLOWED_LEFT_OPERAND_TYPES: Valuetype[] = [
     PrimitiveValuetypes.Text,
     PrimitiveValuetypes.Integer,
     PrimitiveValuetypes.Decimal,
   ];
+  private readonly ALLOWED_RIGHT_OPERAND_TYPES: CollectionValuetype[] =
+    this.ALLOWED_LEFT_OPERAND_TYPES.map((v) => new CollectionValuetype(v));
 
   computeType(
     leftOperandType: Valuetype,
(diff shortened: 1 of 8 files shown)


libs/language-server/src/lib/ast/expressions/type-computers/in-operator-type-computer.ts
```diff
@@ -0,0 +1,79 @@
...
+      return undefined;
+    }
+    assert(
+      isCollectionValuetype(rightOperandType, PrimitiveValuetypes.Decimal) ||
+        isCollectionValuetype(rightOperandType, PrimitiveValuetypes.Integer) ||
+        isCollectionValuetype(rightOperandType, PrimitiveValuetypes.Text),
+    );
+
+    if (
+      !leftOperandType.isConvertibleTo(rightOperandType.elementType) &&
```
rhazn, line 66: Leaving this for reference, in our walkthrough we talked about this not being needed. If either operand is convertible without loss of precision to the other we can use the `in` operator.

rhazn (approved):
Reviewed in walkthrough.

libs/language-server/src/lib/ast/expressions/type-computers/in-operator-type-computer.ts
georg-schwarz: Addressed in [18ac91c](https://github.com/jvalue/jayvee/pull/363/commits/18ac91c8564ff5b7cc36190f17449a6b48450321)

The implications are the following (by example):
- `2.3 in [3]` => not allowed
- `2.0 in [3]` => allowed as 2.0 is interpreted as integer
- `2 in [3.2]` => allowed as 2 can be converted as decimal
- `2.5 in [3.0] => not allowed as 3.0 is interpreted as integer

The last case might not be desired IMO. Will investigate on how to fix it.

georg-schwarz: We use `Number.isInteger` to infer whether a number is an integer or a decimal. The best solution IMO would be to introduce a `UnionType` that allows to say "is oftype x or oftype y". However, this is out of scope of this PR.

I think with the current implementation, it is quite hard to fix since the parsing is done way beforehand and I can only access the number value...

rhazn: Imho should be just an `or`, not an `and`. That way if either side is convertible in the other it is fine. We can try converting the needle into the haystack type by default, if that does not work convert every value in the haystack in the needle type.

georg-schwarz: That's not implementable right now without a larger refactoring.

Thus, I removed the inconsistency that numbers like `3.0` were interpreted as `integer`. With the last change, every number with a dot and number digits is interpreted as a decimal, without always as an integer.

The long-term solution could be to introduce a `UnionValuetype` that interprets such ambiguous numbers as `integer or decimal`. I'd like to tackle this in a separate PR, though.

georg-schwarz: I continued playing around with it and reverted the changes we introduced. This leads to the following state we try to convert the right operand type to the left operand type. Please note that it is not possible to go by values but only by the general type.

- `2.3 in [3]` => allowed as 3 (right) can be converted to a decimal (left)
- `2.0 in [3]` => allowed as 2.0 (left) is interpreted as integer (right)
- `2 in [3.2]` => allowed as 2 (left) can be converted as decimal (right)
- `2.5 in [3.0]` => allowed as 3.0 (right) is interpreted as integer but can be converted to a decimal (left)

In terms of evaluation, we have the following special cases with this approach:
- `2 in [2.0]` => true
- `2.0 in [2]` => true
- `2.00000000000000000000000000000001 in [2]` => true due to rounding losses
- `2 in [2.00000000000000000000000000000001]` => true due to rounding losses

Would that be alright with you? @rhazn

rhazn: Very confusing description of the "allowing". I am fine with all your examples, but I think the messaging/thinking about it is unclear. We should have very simple, clear rules (and I think they are very easy to state):

- For the in operator we have a type for the needle and a type for every element in the haystack
- We have automated type conversion for some types
- We allow automated type conversion as long as it is lossless and clearly defined (integer to decimal as of now)
- We allow any combination that has either: An automated type conversion from needle type to the type of every element in the haystack or the other way around
- For execution, we always try to convert the needle first, if that does not work we convert every element in the haystack, afterwards we compare

For evaluation, it's unfortunate but something we should fix outside of this PR with #305

georg-schwarz: > Very confusing description of the "allowing".

Thanks for pointing that out. Being precise is important! With "allow", i mean accepting it as an input passing the validations, which is separate from the execution/interpretation.

> We should have very simple, clear rules.

Love it! I agree with all of them, but would rephrase these ones:

> - We allow any combination that has either: An automated type conversion from needle type to the type of every element in the haystack or the other way around.

Currently, we track the type of the whole haystack as a combined collection type. The elements of a collection adhere to a nested type for the elements. So here, for example: `CollectionValuetype<IntegerValuetype>` or `CollectionValuetype<DecimalValuetype>` (depending on the example). What might lead to confusion is the wording "type of every element", indicating that really go element by element instead of assessing the collections inner type.

Thus, I'd rephrase this rule to:
- We allow any combination that has either: An automated type conversion from needle type to the type of the elements in the haystack or the other way around.


> - For execution, we always try to convert the needle first, if that does not work we convert every element in the haystack, afterwards we compare

I'd leave that out as we currently don't convert anything for execution as we use TypeScript's joint number type.

rhazn: Sounds good, the updated rules:

- For the in operator we have a type for the needle and a type for every element in the haystack.
- We have automated type conversion for some types.
- We allow automated type conversion as long as it is lossless and clearly defined (integer to decimal as of now).
- We allow any combination that has either: An automated type conversion from needle type to the type of the elements in the haystack or the other way around.

They were also more meant for us to have a shared understanding. Conceptually I think my execution comment still makes sense, but with our current implementation we might not need it yeah.

In any case, I am fine with this.

georg-schwarz: [6fae708](https://github.com/jvalue/jayvee/pull/363/commits/6fae708d2012441bd30ab11f83384114ec64881e) documents this in the docs.

PR #369, opened 2023-06-22
Block example validation

f3l1x98:
This PR adds tests that validate the examples defined inside the BlockMetaInformation of all the blocks defined inside the stdLangExtension.
Additionally it fixes all issues found due to this test.


libs/extensions/std/lang/src/meta-inf-example-validation.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+      const metaInf = new metaInfClass();
+      return [metaInf.type, metaInf];
+    }),
+  )(
+    'should have no error on %s example validation',
+    async (type, blockMetaInf) => {
+      for (const example of blockMetaInf.docs.examples ?? []) {
+        const text = await generateFullJvExample(blockMetaInf, example.code);
+
+        // Validate input
```
georg-schwarz, line 112: Unnecessary comment spotted ;-)

libs/extensions/std/lang/src/meta-inf-example-validation.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+        const validationResult = await validate(text);
+        const diagnostics = validationResult.diagnostics;
+        // Expect 0 errors
```
georg-schwarz, line 115: Unnecessary comment spotted ;-)

libs/extensions/std/lang/src/meta-inf-example-validation.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+    options?: ParseHelperOptions,
+  ) => Promise<LangiumDocument<AstNode>>;
+  let validate: (input: string) => Promise<ValidationResult<AstNode>>;
+
+  let locator: AstNodeLocator;
+
+  const testExtension = new TestLangExtension();
+  const stdExtension = new StdLangExtension();
+
+  async function generateFullJvExample(
```
georg-schwarz, line 39: Can we pull this out from the `describe` to the bottom of the file?

libs/extensions/std/lang/src/meta-inf-example-validation.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+    blockMetaInf: BlockMetaInformation,
+    blockExample: string,
+  ): Promise<string> {
+    // Get name of block in example
```
georg-schwarz, line 43: These kind of comments always indicate that the logic could be extracted into a well-named method instead.

libs/extensions/std/lang/src/meta-inf-example-validation.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+    expectNoParserAndLexerErrors(parsedExample);
+    const blockName: string = (
+      locator.getAstNode<BlockDefinition>(
+        parsedExample.parseResult.value,
+        'pipelines@0/blocks@0',
+      ) as BlockDefinition
+    ).name;
+
+    let pipelineContent = `${blockExample}`;
+    // Generate extractor block and pipe
```
georg-schwarz, line 54: Indicates extraction to own method instead of comment.

libs/extensions/std/lang/src/meta-inf-example-validation.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+        blockMetaInf.inputType,
+        'output',
+      );
+      // Generate block
+      const extractorBlock = `block TestExtractorBlock oftype ${inputBlockMetaInf.type} {}`;
+      // generate pipe
+      const pipe = `TestExtractorBlock -> ${blockName};`;
+      pipelineContent += `\n${extractorBlock}\n${pipe}`;
+    }
+    // Generate loader block and pipe
```
georg-schwarz, line 68: Indicates extraction to own method instead of comment.

georg-schwarz (approved):
LGTM!

Left some minor comments on clean code suggestions

libs/extensions/std/lang/src/meta-inf-example-validation.spec.ts
f3l1x98: Yes I could extract it into own method, but honestly this will never be used anywhere else and I personally do not like extracting just for the sake of extracting because it results into a lot of jumping through files when trying to understand the code.  
But if you really prefer me extracting this (and the following parts) into separate methods I will do so.

libs/extensions/std/lang/src/meta-inf-example-validation.spec.ts
f3l1x98: Only if I extend the parameters passed to the function to also include most of the in `beforeAll` initialized variables.
``` js
async function generateFullJvExample(
  blockMetaInf: BlockMetaInformation,
  blockExample: string,
): Promise<string> {}
```
would become
``` js
async function generateFullJvExample(
  blockMetaInf: BlockMetaInformation,
  blockExample: string,
  parse: (
    input: string,
    options?: ParseHelperOptions,
  ) => Promise<LangiumDocument<AstNode>>,
  locator: AstNodeLocator,
  testExtension: TestLangExtension,
): Promise<string> {}
```

libs/extensions/std/lang/src/meta-inf-example-validation.spec.ts
georg-schwarz: Yes, I think it contributes to better code readability. Keeping methods short and understandable by good method names is good step into the direction of clean code.

You can extract them to the same file, so you'd at least stay in the same file when jumping through the code ;-)

libs/extensions/std/lang/src/meta-inf-example-validation.spec.ts
georg-schwarz: Ah well, then I'd just leave it.

PR #377, opened 2023-06-25
Constraint example validation test

f3l1x98:
This PR adds a test that validates the examples of all `ConstraintMetaInformation` that are part of the language-server.

Note: I had to add `getConstraintMetaInf()` inside `constraint-registry.ts` because the existing methods (like `getRegisteredConstraintMetaInformation`) did not work. The reason for that is that `it.each` gets resolved before `beforeAll`, but the population of the meta-inf-registry happens inside `beforeAll`.


libs/language-server/src/lib/constraint/constraint-registry.ts
```diff
@@ -2,18 +2,28 @@
...
 import { registerMetaInformation } from '../meta-information/meta-inf-registry';
+import { ConstructorClass } from '../util';
 
 import { AllowlistConstraintMetaInformation } from './allowlist-constraint-meta-inf';
 import { DenylistConstraintMetaInformation } from './denylist-constraint-meta-inf';
 import { LengthConstraintMetaInformation } from './length-constraint-meta-inf';
 import { RangeConstraintMetaInformation } from './range-constraint-meta-inf';
 import { RegexConstraintMetaInformation } from './regex-constraint-meta-inf';
 
+export function getConstraintMetaInf(): ConstructorClass<ConstraintMetaInformation>[] {
```
georg-schwarz, line 15: Consider to export a constant `DefaultConstraintMetaInfs` or so instead.

georg-schwarz (approved):
lgtm

libs/language-server/src/lib/constraint/constraint-registry.ts
f3l1x98: I thought about it, but choose this design in order to be more consistent with `JayveeLangExtension#getBlockMetaInf`

PR #383, opened 2023-06-28
Target specific blocks for debug logging

georg-schwarz:


apps/interpreter/src/interpreter.ts
```diff
@@ -299,3 +304,14 @@ export function logPipelineOverview(
...
+
+function getDebugTargets(
+  debugTargetsString: string | undefined,
+): R.DebugTargets {
+  const areAllBlocksTargeted = debugTargetsString === undefined;
+  if (areAllBlocksTargeted) {
+    return 'all';
+  }
+
+  return debugTargetsString.split(',');
```
rhazn, line 316: Should probably also include removing whitespaces (`trim()`).

libs/execution/src/lib/blocks/block-executor.ts
```diff
@@ -52,6 +54,15 @@ export abstract class AbstractBlockExecutor<I extends IOType, O extends IOType>
       return;
     }
 
+    const currentNode = context.getCurrentNode();
+    assert(isBlockDefinition(currentNode));
+    const isBlockTargeted =
```
rhazn, line 59: This should be a function somewhere (on the DebugLogVisitor (?)) so it is easier to change in the future.

libs/execution/src/lib/debugging/debug-configuration.ts
```diff
@@ -0,0 +1,11 @@
+// SPDX-FileCopyrightText: 2023 Friedrich-Alexander-Universitat Erlangen-Nurnberg
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+
+export const DebugGranularityValues = ['peek', 'exhaustive', 'skip'] as const; // convention: last item is the default value
```
rhazn, line 5: Conventions are terrible, can we just define it as a default?

rhazn (approved):

apps/docs/docs/user/intro.md
```diff
@@ -47,7 +47,16 @@ Run with **additional debug output**:
...
+
+With **runtime parameters**:
+
+```console
+jv -e <param>=<value> -e <param>=<value> ... <file>
+```
+
+### Debug a `.jv` file
+
+Currently, we only allow print debugging by the parameters `dg` and `dt`.
```
rhazn, line 59:
```suggestion
Print debugging is controlled by the parameters `--debug-granularity` and `--debug-target`.
```

rhazn: Reading this again I think we should always use the long form parameters in the docs, that way they are easier to remember and understand.

Also, I'd never use 'currently' or 'only' in the docs. We can have a roadmap with plans but the docs are a place to describe the current state.

rhazn: Also, reading this I was initially confused if using `-d` now logs nothing because `skip` is the default for granularity. If I understand right, `skip` in that context means "Skip extra debug output with data for each block". But it reads really confusing. I'd rename `--debug-granularity` to something like `--debug-data` or `skip` to `pipeline-only` (not the best name but the only thing I spontaneously got).

rhazn (approved):

apps/docs/docs/user/intro.md
georg-schwarz: Thx. I agree that naming is quite hard here...
How about `debug-output-granularity` (`dg`) and `debug-output-targets` (`dt`)?

rhazn: If you have a `debug` flag and call anything a debug granularity of `skip`, I would assume it to skip debug output. I would honestly just rename `skip` to something like `minimal`?

PR #384, opened 2023-06-28
Block custom validation tests

f3l1x98:
This PR adds some tests for testing the custom validation of all blocks (which have custom validation logic) in the std-extension and tabular-extension (there are none in the rdbms-extension)


libs/extensions/std/lang/src/gtfs-rt-interpreter-meta-inf.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    const validationResult = await validate(text);
+    const diagnostics = validationResult.diagnostics;
+
+    expect(diagnostics).toHaveLength(0);
+  });
+
+  it('should diagnose error on invalid entity parameter value', async () => {
+    const text = readJvTestAsset(
+      'gtfs-rt-interpreter-meta-inf/invalid-invalid-entity-param.jv',
```
georg-schwarz, line 49: Duplicate `invalid` in file name. At other places as well ;-)

libs/extensions/std/lang/src/text-file-interpreter-meta-inf.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+    useExtension(new StdLangExtension());
+    // Register test extension
+    useExtension(new TestLangExtension());
+    // Create language services
+    const services = createJayveeServices(NodeFileSystem).Jayvee;
+    // Create validation helper for language services
+    validate = validationHelper(services);
+  });
+
+  it('should diagnose error on invalid encoding parameter value', async () => {
```
georg-schwarz, line 36: Does it make sense to add one with a valid encoding parameter?

georg-schwarz: Or for all valid ones with `it.each`?

libs/extensions/std/lang/src/text-line-deleter-meta-inf.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+    useExtension(new StdLangExtension());
+    // Register test extension
+    useExtension(new TestLangExtension());
+    // Create language services
+    const services = createJayveeServices(NodeFileSystem).Jayvee;
+    // Create validation helper for language services
+    validate = validationHelper(services);
+  });
+
+  it('should diagnose error on line parameter less or equal to zero', async () => {
```
georg-schwarz, line 36: Does it make sense to add a positive test as well?

libs/extensions/std/lang/src/text-range-selector-meta-inf.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+    useExtension(new StdLangExtension());
+    // Register test extension
+    useExtension(new TestLangExtension());
+    // Create language services
+    const services = createJayveeServices(NodeFileSystem).Jayvee;
+    // Create validation helper for language services
+    validate = validationHelper(services);
+  });
+
+  it('should diagnose error on lineFrom parameter less or equal to zero', async () => {
```
georg-schwarz, line 36: Add positive test (?)

libs/extensions/tabular/lang/src/lib/cell-writer-meta-inf.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+    expect(diagnostics).toEqual(
+      expect.arrayContaining([
+        expect.objectContaining({
+          message: 'The cell range needs to be one-dimensional',
+        }),
+      ]),
+    );
+  });
+
+  it('should diagnose error on number of write values does not match cell range', async () => {
```
georg-schwarz, line 57: add positive test(?)

georg-schwarz (approved):
Awesome! Looks pretty good to me!

I'm asking myself if we should ask some positive case tests as well instead only negative ones? Might be useful to have them as regression tests.

libs/extensions/std/lang/src/gtfs-rt-interpreter-meta-inf.spec.ts
f3l1x98: That naming resulted from my convention that every invalid jv file starts with `invalid` and I couldn't come up with anything better to describe this file than `invalid-entity-param`

libs/extensions/std/lang/src/text-file-interpreter-meta-inf.spec.ts
f3l1x98: Considering that the code that directly determines whether the encoding is valid wasn't written by us I think that would not be necessary.

georg-schwarz (approved):

PR #386, opened 2023-06-29
Constraint example validation test

f3l1x98:
This PR adds a test that validates the examples of all `ConstraintMetaInformation` that are part of the language-server.

Note: I had to add `getConstraintMetaInf()` inside `constraint-registry.ts` because the existing methods (like `getRegisteredConstraintMetaInformation`) did not work. The reason for that is that `it.each` gets resolved before `beforeAll`, but the population of the meta-inf-registry happens inside `beforeAll`.


f3l1x98:
Redone #377

PR #388, opened 2023-07-05
Interpreter as a library

georg-schwarz:
These changes extract most of the logic of the `interpreter` to a library `interpreter-lib`. This is now reusable for runtimes to depend on instead of spawning a subprocess.

It is yet not clear whether some functionality should flow back into the interpreter. Time will tell, I guess.

We should also consider renaming the `interpreter` to something like `cli` and rename the `interpreter-lib` to `interpreter`. But I'd like to hear your opinion first @rhazn since it might cause confusion if the `interpreter` suddenly becomes something else.


libs/interpreter-lib/README.md
```diff
@@ -0,0 +1,7 @@
+<!--
+SPDX-FileCopyrightText: 2023 Friedrich-Alexander-Universitat Erlangen-Nurnberg
+
+SPDX-License-Identifier: AGPL-3.0-only
+-->
+
+# Jayvee interpreter Library
```
rhazn, line 7: This could be expanded a bit, maybe even with a minimal mermaidjs diagram that shows how the CLI/the interpreter work together.

libs/interpreter-lib/project.json
```diff
@@ -0,0 +1,59 @@
...
+  "name": "interpreter-lib",
+  "$schema": "../../node_modules/nx/schemas/project-schema.json",
+  "sourceRoot": "libs/interpreter-lib/src",
+  "projectType": "library",
+  "targets": {
+    "build": {
+      "executor": "@nrwl/js:tsc",
+      "outputs": ["{options.outputPath}"],
+      "options": {
+        "external": "none", // bundle dependencies to make usable externally
```
rhazn, line 11: JSON does not know comments.

rhazn (approved):
I think renaming `interpreter` to `jayvee-cli`/`interpreter-cli` would be good. On the other hand, I like keeping `interpreter-lib`.

libs/interpreter-lib/README.md
georg-schwarz: Done in [43b6910](https://github.com/jvalue/jayvee/pull/388/commits/43b6910fd787c25947adab4ac9e5e8b75df70750)

libs/interpreter-lib/project.json
georg-schwarz: Done in [f76ae1f](https://github.com/jvalue/jayvee/pull/388/commits/f76ae1f472c299da41d4352532f9100f31b7a0aa)

PR #404, opened 2023-07-19
Implement Retry Mechanism for HttpExtractor

### LINKED ISSUE #403: [FEATURE] Retry Mechanism for HttpExtractor
## User Story
1. As a {Jayvee user}
2. I want {to configure retries on the HTTP executor}
3. So that {I can build a more robust pipeline}

## User Acceptance Criteria
- [x] There is a property `retries` defining how many retries should be executed on failure
- [x] There is a property `retryBackoff` defining how long should be waited before a retry
- [x] Documentation is updated

## Definitions of Done
- [x] A PR has been opened and accepted
- [x] All user acceptance criteria are met
- [x] All tests are passing

georg-schwarz:
Closes #403.

## Example
```
		url: "https://gist.githubusercontent.com/noamross/e5dasdasd..."; // invalid URL
		retries: 3;
		retryBackoff: 10000;
	}

```
image

### DIFF
--- a/libs/extensions/std/exec/src/http-extractor-executor.ts
+++ b/libs/extensions/std/exec/src/http-extractor-executor.ts
@@ -51,7 +51,7 @@ export class HttpExtractorExecutor extends AbstractBlockExecutor<
       PrimitiveValuetypes.Integer,
     );
     const retryBackoff = context.getPropertyValue(
-      'retryBackoff',
+      'retryBackoffMilliseconds',
       PrimitiveValuetypes.Integer,
     );
 
--- a/libs/extensions/std/exec/src/http-extractor-executor.ts
+++ b/libs/extensions/std/exec/src/http-extractor-executor.ts
@@ -28,8 +28,8 @@ import {
   inferMimeTypeFromContentTypeString,
 } from './file-util';
 import {
-  BackoffStrategy,
-  LinearBackoffStrategy,
+  createBackoffStrategy,
+  isBackoffStrategyHandle,
 } from './util/backoff-strategy';
 
 type HttpGetFunction = typeof http.get;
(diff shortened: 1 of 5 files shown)


libs/extensions/std/exec/src/http-extractor-executor.ts
```diff
@@ -43,14 +46,39 @@ export class HttpExtractorExecutor extends AbstractBlockExecutor<
...
-    const file = await this.fetchRawDataAsFile(url, context);
-
-    if (R.isErr(file)) {
-      return file;
+    const retries = context.getPropertyValue(
+      'retries',
+      PrimitiveValuetypes.Integer,
+    );
+    const retryBackoff = context.getPropertyValue(
+      'retryBackoff',
```
rhazn, line 54: I am a big fan in including expected units in variable names that are not obvious. Id suggest `retryBackoffMS` or better `retryBackoffMicroseconds`.

rhazn (changes requested):
I think our default strategy should be exponential backoffs and we should enforce some form of minimal delay. Right now it is trivial to missconfigure (or deliberately configure in the case of JV cloud) this to spam a source with 1ms delayed requests.

I'd go with exponential backoff and minimal delay of 2s. So we'd have 2, 4, 8... etc backoffs.

libs/extensions/std/exec/src/http-extractor-executor.ts
georg-schwarz: With the comment about the exponential backoff strategy, does it make sense to remove this prop completely?

rhazn: We could also go with just exponential backoffs, 1, 2, 4 etc seconds. But might as well keep both configurable, right? Might be useful if we add a prop to change backoff strategy in the future.

georg-schwarz:
- Added a prop `retryBackoffStrategy` with allowed values `exponential` (default) and `linear`.
- Renamed `retryBackoff` to `retryBackoffMilliseconds`
- Prop `retryBackoffMilliseconds` must be 1000 or larger

rhazn (approved):
It is wrong but intuitively, I'd assume exp backoff to be 1, 2, 4... for 1000ms. Maybe we force a min value of 2000 or do the 1,2,4 by hand?

I am fine with merging it like this as well though.

PR #405, opened 2023-07-21
feature/#392_Unpack_gz_files_and_#391follow_redirects

### LINKED ISSUE #391: [FEATURE] HTTPExtractor should follow redirects
## User Story
1. As a {user/persona}
2. I want {that the HTTPExtractor can download files from S3 buckets}
3. So that {I can use data in S3}

## User Acceptance Criteria
- [ ] A new optional property "followRedirects" exists for the HTTPExtractor block, by default it is true
- [ ] if followRedirects is true, HTTPExtractor follows redirects

## Examples
- Example file: https://figshare.com/ndownloader/files/33982973

## Notes
- https://jvalue.github.io/jayvee/docs/user/block-types/HttpExtractor
- https://jvalue.github.io/jayvee/docs/dev/intro

## Definitions of Done
- [ ] All tests are passing

### LINKED ISSUE #392: [FEATURE] Unpack gz files
## User Story
1. As a {user/persona}
2. I want {to be able to unpack gz files}
3. So that {I can use them}

## User Acceptance Criteria
- [ ] ArchiveInterpreter is able to handle ".gz" files

## Examples
- example file (wait for #391 ) https://figshare.com/ndownloader/files/28333845

## Notes
- https://jvalue.github.io/jayvee/docs/user/block-types/ArchiveInterpreter
- https://jvalue.github.io/jayvee/docs/dev/intro

## Definitions of Done
- [ ] All tests are passing

A-M-A-X:
# [FEATURE] HTTPExtractor should follow redirects #391
- Added 'gz' file-type to `Archive-Interpreter`
  - Implemented with zlib 
  - Added Testfile `thermoelectricMaterials.jv`
     - only showcases that `File-Picker` receives a file - error on csv interpreter, as underlying file is in json format
Closes #392 

# [FEATURE]  Unpack gz files #392 
- Added redirect logic to `HTTP-Extractor`
  - Used [follow-redirects](https://www.npmjs.com/package/follow-redirects) -package
  - Used default number of redirects from package (`max redirects = 21`)
  - `followRedirects `attribute in HTTP-Extractor defaults to `true`
  - Added Testfile `materials.jv`
Closes #391


example/thermoelectricMaterials.jv
rhazn, line 1: I don't think we need new examples in the Jayvee repository right now, it would be better to move this to the SWC repo.

Also please don't provide examples that do not work. These files are meant as documentation for users of Jayvee, not to test implementation ;).

example/materials.jv
rhazn, line 1: I don't think we need new examples in the Jayvee repository right now, it would be better to move this to the SWC repo.

georg-schwarz:
@f3l1x98 can you provide guidance on how the failing example file can be turned into a unit test?

libs/extensions/std/exec/src/archive-interpreter-executor.ts
```diff
@@ -55,12 +56,69 @@ export class ArchiveInterpreterExecutor extends AbstractBlockExecutor<
...
+      return R.ok(fs.right);
+    }
+
     return R.err({
       message: `Archive is not a zip-archive`,
       diagnostic: { node: context.getCurrentNode(), property: 'name' },
     });
   }
 
+  private async loadGzFileToInMemoryFileSystem(
```
rhazn, line 76: This function should duplicate a lot of the `loadZipFileToInMemoryFileSystem` logic. I assume only the extraction of files would be different, putting them into our in memory filesystem etc is all repeated logic.

It would be better to implement it that way, e.g., by extracting the shared logic of putting files into our in-memory file system into one function and reduce these two functions to just extracting the relevant archives and then using the new function to put files into a file system.

libs/extensions/std/exec/src/archive-interpreter-executor.ts
```diff
@@ -55,12 +56,69 @@ export class ArchiveInterpreterExecutor extends AbstractBlockExecutor<
...
+        inferFileExtensionFromFileExtensionString(extName) ||
+        FileExtension.NONE;
+      const file = new BinaryFile(
+        fileName,
+        fileExtension,
+        mimeType,
+        archivedObject,
+      );
+      
+      const addedFile = fs.putFile(
```
rhazn, line 103: I am very confused how this works because I can only see one `putFile` call and no loops or iterations. Does this actually correctly expand gz archives with multiple files in them?

libs/extensions/std/lang/src/archive-interpreter-meta-inf.ts
```diff
@@ -19,7 +19,7 @@ export class ArchiveInterpreterMetaInformation extends BlockMetaInformation {
         archiveType: {
           type: PrimitiveValuetypes.Text,
           docs: {
-            description: 'The archive type to be interpreted, e.g., `"zip"`.',
+            description: 'The archive type to be interpreted, e.g., `"zip" or "gz`.',
```
rhazn, line 22:
```suggestion
            description: 'The archive type to be interpreted, e.g., `"zip" or "gz".',
```

libs/extensions/std/lang/src/http-extractor-meta-inf.ts
```diff
@@ -147,6 +147,19 @@ export class HttpExtractorMetaInformation extends BlockMetaInformation {
             }
           },
         },
+        followRedirects: {
+          type: PrimitiveValuetypes.Boolean,
+          defaultValue: true,
+          docs: {
+            description: 'Indicates, whether to follow redirects on get requests. If `false`, Redirects are disabled. Default `true`',
```
rhazn, line 154: Redirects are not disabled, just not followed ;).

```suggestion
            description: 'Indicates, whether to follow redirects on get requests. If `false`, redirects are not followed. Default `true`',
```

libs/extensions/std/exec/src/http-extractor-executor.ts
```diff
@@ -126,6 +126,17 @@ export class HttpExtractorExecutor extends AbstractBlockExecutor<
           );
         }
 
+        if (responseCode === 302) {
```
rhazn, line 129: 301 is also a redirect code I assume.

rhazn (commented):
I know you did not tag me in this to review but it popped up in my email. I left some spontaneous commments. For future work, it would be good to have only one PR per issue please.

libs/extensions/std/exec/src/archive-interpreter-executor.ts
A-M-A-X: To my understanding gzip is used to compress individual files. To compress multiple files with gzip, they have to be concated first, e.g to a .tar file. This I would implement in a different feature. Happy to further look into this, if you disagree.

libs/extensions/std/lang/src/http-extractor-meta-inf.ts
A-M-A-X: Changed as suggested

libs/extensions/std/lang/src/archive-interpreter-meta-inf.ts
A-M-A-X: Changed to 'The archive type to be interpreted, e.g., "zip" or "gz".',

libs/extensions/std/exec/src/http-extractor-executor.ts
A-M-A-X: Changed to >= 301, as 300 is used for 'multiple choices'. The remaining `300 < response codes < 400` point to redirect location or are not used in get context.

A-M-A-X: ok, done

f3l1x98:
> @f3l1x98 can you provide guidance on how the failing example file can be turned into a unit test?

If help is needed/wanted, yes.
Though we currently do not have any tests yet, where this could be tested.
I think this would be part of either the test for the `CSVInterpreterBlockExecutor` or End-to-End tests, but would have to look into it.

rhazn:
@A-M-A-X (in case you want me to review this again, you'll need to re-request the review on the top right, little icon next to my name).

libs/extensions/std/exec/src/archive-interpreter-executor.ts
```diff
@@ -133,36 +100,47 @@ export class ArchiveInterpreterExecutor extends AbstractBlockExecutor<
...
-        message: `Unexpected Error ${
-          error instanceof Error ? error.message : JSON.stringify(err)
-        } occured during processing`,
-        diagnostic: { node: context.getCurrentNode(), property: 'name' },
-      });
+      return R.err(this.generateErrorObject(context, error));
     }
   }
+
+  private createBinaryAndPutFile(
```
rhazn, line 114: I think there are two problems with this:

- Methods that have an "and" in their name are almost always to large and can be split up in an obvious way ;). In this case, I'd split this into `createBinary` and `putFile` theoretically.
- Changing objects that are parameters to a function as a sideeffect is often confusing and leads to errors. Here I would not change the inmemoryfilesystem as sideeffect of the function.

Combined, I'd rename the method to `createFileFromArchive` (or similar) and just have line 120 to 128 in it, return the new `BinaryFile`. Then habe the fs.putfile and assert outside of the function on it's own, I don't think they need their own function.

rhazn (changes requested):
Small changes to the function still but otherwise looks good 👍

libs/extensions/std/exec/src/archive-interpreter-executor.ts
A-M-A-X: - Moved `fs.putFile` and assertion outside of function. 
- Renamed function to `createFileFromArchive` and have it return the new `BinaryFile`. 
- Removed now unused parameters.

rhazn (approved):

rhazn:
Wait

rhazn:
Ah, nvm I am blind. It's fine, you can merge this into dev.

A-M-A-X:
@rhazn I think I am missing write acess to do the merge myself, or am I blind as well?

rhazn:
Ah yeah might be on me. It's fine, I can merge these finally.

PR #406, opened 2023-07-21
Builtin Valuetypes

georg-schwarz:
Introduces `builtin valuetype`s to the language (replaces former keywords).

Not sure if we really need this, it was majorly for me learning on the job for the builtin blocks. However, users can now click on the builtin valuetypes like `text`, `decimal`, `integer`, and `boolean` and see their definition in VSCode. This would allow better in-code documentation as we can add comments on the behavior of the builting valuetypes in the standard library.

[Screencast from 07-21-2023 02:53:56 PM.webm](https://github.com/jvalue/jayvee/assets/28054628/317eb37e-7a62-402d-93ae-3744584b9e6b)

Let me know what you think! If you think it is useful I'd introduce some comments to the builtin valuetypes.

### DIFF
--- a/apps/docs/docs/user/.gitignore
+++ b/apps/docs/docs/user/.gitignore
@@ -1,0 +1,1 @@
+builtin-valuetypes.md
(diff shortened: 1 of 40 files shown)


rhazn (approved):
LGTM

PR #409, opened 2023-07-29
[RFC] for single-attribute value types

dirkriehle:
Adds alternative syntax to current single-attribute value types and prepares the way for multi-attribute value types.


dirkriehle:
I have read the CLA Document and I hereby sign the CLA

rfc/0014-attribute-based-valuetypes/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+This syntax is smart in that you don't have to list and name an attribute but rather rely on an implicit 'value' attribute. Still, I propose to make that attribute explicit. New syntax would be:
+
+```jayvee
+valuetype CorrelationCoefficient {
+  attributes: [
+    value oftype decimal;
+  ];
+  constraints: [
+    MinusOneToPlusOneRange; // Don't know how to attach this to value attribute
```
georg-schwarz, line 53: If we refactor the constraints, the constraints could always refer to the complete valuetype (atomic or composite). This would also allow constraints spanning different attributes of a valuetype.

Probably bad example with the alternative flat syntax I proposed in another comment:

```
valuetype SumIsOne {
    attribute x1 oftype decimal;
    attribute x2 oftype decimal;

    constraint sumIsOne: x1 + x2 == 1.0;
    constraint x1isPositive: x1 > 0;
    constraint x2isPositive: x2 > 0;
}
```

georg-schwarz: Alternative, with the current constraints, we could introduce sth like this `on` keyword:

```
valuetype CorrelationCoefficient {
    attributes: [
        value oftype decimal,
    ];
    constraints: [
        MinusOneToPlusOneRange on value,
    ]
}
```

rhazn:
recheck

rfc/0014-attribute-based-valuetypes/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Drawbacks
+
+1. Introduces a redundant syntax,
+2. Creates extra work/may be too difficult, and
+3. May be disruptive to how the language currently works. 
+
+## Alternatives
+
+Conceivably, we could use multiple inheritance for multi-attribute value types... just joking.
```
georg-schwarz, line 84: But how about inheritance in general? 

Right now, this syntax does not allow an `extension` of an existing valuetype. I would see value offering it, though:

```
valuetype PositiveCorrelationCoefficient extends CorrelationCoefficient {
    // inherits all attributes
    // could add further attributes?

    // inherits all constraints
    // could ad further constraints
    constraints: [
         ZeroToPlusOneRange,
    ]
}
```

georg-schwarz (commented):
Love it!

I think we should consider adapting the constraints syntax to be better suited for the extension to multi-attribute valuetypes (see the comments).

rfc/0014-attribute-based-valuetypes/README.md
```diff
@@ -0,0 +1,88 @@
...
+```
+
+While more verbose, it prepares the way for 
+
+```jayvee
+valuetype Money {
+  attributes: [
+    amount oftype decimal;
+    currency oftype Currency;
+  ];
```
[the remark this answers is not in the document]
rhazn: We've used `property` instead of attributes in built-in blocks (https://github.com/jvalue/jayvee/tree/main/rfc/0008-builtin-blocktypes). I'd use the same naming here. For blocks, we decided properties are values an object owns intrinsically that you use in modeling, attributes is something you attribute to objects from the outside. The distinction is very faint but I would argue for consistency and reuse `property` here.

Otherwise I think the syntax is nice and would go for:

```
valuetype Money {
  property amount oftype decimal;
  property currency oftype Currency;
}
```

rfc/0014-attribute-based-valuetypes/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+<!-- 
+  TODO: Explain the details of the RFC. 
+  If the RFC contains more than a single cohesive aspect, structure this section accordingly.
+  Make sure to provide realistic modelling examples on the example data set introduced above.
+-->
+
+## Drawbacks
+
+1. Introduces a redundant syntax,
```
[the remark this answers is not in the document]
rhazn: I'd also replace the syntax.

rfc/0014-attribute-based-valuetypes/README.md
rhazn: I think the first flat syntax is nice, it could work with references constraints by instantiating them and assigning the values from attributes from them. 

Consider:

```
constraint IsPositive on integer: value >= 0;

valuetype SumIsOne {
    attribute x1 oftype decimal;
    attribute x2 oftype decimal;

    constraint sumIsOne: x1 + x2 == 1.0;
    constraint x1isPositive: IsPositive(value = x1);
    constraint x2isPositive: IsPositive(value = x2);
}
```

This is a pretty big change because it introduces syntax for instantiating a constraint and assigning values to it but I'd like to throw it out there for discussion. Some syntax alternatives for that are possible of course, but I also like explicit parameter naming.

rfc/0014-attribute-based-valuetypes/README.md
rhazn: I'd vote for keeping inheritance out of this RFC to have a smaller scope :).

rfc/0014-attribute-based-valuetypes/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+```jayvee
+valuetype CorrelationCoefficient oftype decimal {
+  constraints: [
+    MinusOneToPlusOneRange,
+  ];
+}
+```
+
+This syntax is smart in that you don't have to list and name an attribute but rather rely on an implicit 'value' attribute. Still, I propose to make that attribute explicit. New syntax would be:
```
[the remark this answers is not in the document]
rhazn: I don't think we should enforce this, the less naming magic the better. We can assume for a transition period that if a value type has only one attribute, that attribute is mapped to the `value` keyword in expressions for constraints.

rhazn:
> Love it!
> 
> I think we should consider adapting the constraints syntax to be better suited for the extension to multi-attribute valuetypes (see the comments).

Also my thoughts, left some comments as well but I think this is on a good path 👍

rfc/0014-attribute-based-valuetypes/README.md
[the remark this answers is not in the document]
dirkriehle: At first glance, I'm indifferent/don't see a difference between property and attribute. Is there a semantic difference?

Which ever keyword we use, I observe a pattern of singular plural i.e. property/properties and constraint/constraints. 

My take is that users may want both (property for one new property, and properties for a list of properties, constraint for one constraint, and constraints for a set of constraints). If only one is possible, I'd go for singular i.e. Philip's suggestion

rfc/0014-attribute-based-valuetypes/README.md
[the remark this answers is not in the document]
dirkriehle: The beauty of not much legacy yet ;-) Agreed.

rfc/0014-attribute-based-valuetypes/README.md
```diff
@@ -0,0 +1,88 @@
...
+2. Creates extra work/may be too difficult, and
+3. May be disruptive to how the language currently works. 
+
+## Alternatives
+
+Conceivably, we could use multiple inheritance for multi-attribute value types... just joking. 
+
+## Possible Future Changes/Enhancements
+
+This proposal is to prepare the way for multi-attribute value types.
```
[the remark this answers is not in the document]
dirkriehle: We really should put this to the test in AMSE/SAKI/MADE. I have always suspected that beyond very simple uses of generics, folks don't get it/don't use it because it makes their brain jump out of its casing.

rfc/0014-attribute-based-valuetypes/README.md
```diff
@@ -0,0 +1,88 @@
...
+    MinusOneToPlusOneRange,
+  ];
+}
+```
+
+This syntax is smart in that you don't have to list and name an attribute but rather rely on an implicit 'value' attribute. Still, I propose to make that attribute explicit. New syntax would be:
+
+```jayvee
+valuetype CorrelationCoefficient {
+  attributes: [
```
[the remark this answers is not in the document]
dirkriehle: No strong feelings. I used attributes because that's what I know it by. Let me ask ChatGPT for any subtle semantic differences between attribute and property. Try this: "Hi! Is there any significant semantic difference between the words attribute and property when describing something?"

rfc/0014-attribute-based-valuetypes/README.md
dirkriehle: +1 on "flat syntax" over collections (keyword in plural form) if we can have only one.

On Philip's example I observe a subtle syntactic incongruence I think which is the use of : for constraint but not attribute.

rfc/0014-attribute-based-valuetypes/README.md
dirkriehle: I'm all for single inheritance but agree that it is not part of this RFC.

rfc/0014-attribute-based-valuetypes/README.md
rhazn: We had this discussion before I think and it is true. My intuition is that we describe a property of the `SumIsOne` valuetype with both (`attribute x1 oftype decimal` is a property that is of type decimal attribute with the name x1, `constraint sumIsOne` is a property of type constraint with the name sumIsOne). The syntactic difference then comes from assigning a value to one (the constraint has an expression assigned) and assigning no value to the attribute, just declaring it's existence.

rhazn:
I think we have a very similar vision here, our normal flow would be for the RFC creator (in this case you @dirkriehle) to update their RFC with the feedback as the understand it, push it to this PR and retrigger a review.

If you have the time you can follow that workflow, otherwise we'd bring this to a future meeting (e.g., thursday) and can sync in person.

PR #411, opened 2023-08-02
WIP: Composite blocktypes

rhazn:


libs/language-server/src/grammar/blocktype.langium
```diff
@@ -5,18 +5,38 @@
...
 BlocktypeOutput:
     'output' name=ID 'oftype' iotype=[IotypeDefinition] ';';
 
 BlocktypeProperty:
-    'property' name=ID 'oftype' valuetype=ValuetypeReference (':' defaultValue=Expression)? ';';
\ No newline at end of file
+    'property' name=ID 'oftype' valuetype=ValuetypeReference (':' defaultValue=Expression)? ';';
+
+BlocktypePipeline:
+    input=[BlocktypeInput] '->' (blocks+=[BlockDefinition] '->')+ output=[BlocktypeOutput] ';';
```
georg-schwarz, line 42: Just as a remark 'cause I'm unsure if you are aware: This grammar prevents splitting the pipeline into multiple statements like we allow for pipelines:

```
input1 -> IntermediateBlock;
IntermediateBlock -> output1;
```

I believe this "feature" is not required for the first version, but we should have it on the radar.

libs/execution/src/lib/blocks/block-executor-registry.ts
```diff
@@ -22,7 +32,23 @@ export function getRegisteredBlockExecutors(): BlockExecutorClass[] {
 export function createBlockExecutor(block: BlockDefinition): BlockExecutor {
   const blockType = block.type.ref?.name;
   assert(blockType !== undefined);
+
+  if (
+    !blockExecutorRegistry.get(blockType) &&
+    block.type.ref &&
```
georg-schwarz, line 38: I believe this line is unnecessary thank's to line 34.

libs/language-server/src/lib/ast/expressions/type-inference.ts
```diff
@@ -325,9 +327,13 @@ function inferTypeFromReferenceLiteral(
...
-  if (isTransformPortDefinition(referenced)) {
-    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
-    const valueType = referenced?.valueType;
+  if (
+    isTransformPortDefinition(referenced) ||
+    isBlocktypeProperty(referenced)
+  ) {
+    const valueType = isTransformPortDefinition(referenced)
+      ? referenced.valueType
+      : referenced.valuetype;
```
georg-schwarz, line 336: Inconsistency `valuetype` vs. `valueType`. I believe we used camel case for value types in existing code. That should also make a uniform access possible here.

libs/language-server/src/lib/ast/io-type.ts
```diff
@@ -0,0 +1,27 @@
...
+export function getIOType(blockIO: BlocktypeInput | BlocktypeOutput): IOType {
+  const ioName = blockIO.iotype.ref?.name as string;
+
+  assert(
+    Object.values(IOType).some((type) => type === ioName),
+    `IOType ${ioName} does not exist.`,
+  );
+
+  return ioName as IOType;
+}
```
georg-schwarz, line 27: I looked this method up and couldn't find it in the code base on main.
- `ioName` can potentially be undefined I think.
- The conversion to the enum can be a little error prone, but that is just me being very nitpicky here. In case an invalid value survives the assert above (which obviously should not be the case), then the program just continues using the invalid string. With the version below it resolves to `undefined`, making discovering such a bug easier.

```suggestion
export function getIOType(blockIO: BlocktypeInput | BlocktypeOutput): IOType {
  const ioTypeName = blockIO.iotype.ref?.name;
  assert(ioTypeName !== undefined);

  assert(
    Object.values(IOType).some((type) => type === ioTypeName),
    `IOType ${ioTypeName} does not exist.`,
  );

  return IOType[ioTypeName as keyof typeof IOType];
}
```

libs/language-server/src/lib/ast/model-util.ts
```diff
@@ -4,26 +4,47 @@
...
 ): BlockDefinition[] {
+  // For composite blocks the first blocks of all pipelines are starting blocks as they have inputs
+  if (isCompositeBlocktypeDefinition(container)) {
+    const startingBlocks = container.pipes
+      .map((pipe) => pipe.blocks[0])
+      .map((blockRef: Reference<BlockDefinition> | undefined) => {
+        if (
+          blockRef &&
+          blockRef.ref &&
+          getMetaInformation(blockRef.ref.type) !== undefined
```
georg-schwarz, line 33: Can be simplified by using the `?` operator. And I'd be explicit with the `!== undefined` check (just a nitpick).

```suggestion
          blockRef?.ref !== undefined &&
          getMetaInformation(blockRef.ref.type) !== undefined
```

libs/language-server/src/lib/ast/model-util.ts
```diff
@@ -4,26 +4,47 @@
...
+        ) {
+          return blockRef.ref;
+        }
+        return undefined;
+      })
+      .filter(Boolean) as unknown as BlockDefinition[];
```
georg-schwarz, line 39: Just as a hint: this issue is also solvable with a type guard instead, then we can get rid of the type assertion. Type assertions can introduce bugs on refactoring as they overwrite the TypeScript type system, the type guard solution would show an error if the type of `x` was refactored.

```suggestion
      .filter((x): x is BlockDefinition => x !== undefined);
```

libs/language-server/src/lib/meta-information/composite-blocktype-meta-inf.ts
```diff
@@ -0,0 +1,56 @@
...
+      presentPropertyNames,
+    );
+
+    // We assume blocktype properties that have an expression as default value can be evaluated during runtime
+    return missingRequiredPropertyNames.filter((propertyName) => {
+      const blocktypeProperty = this.blockTypeDefinition.properties.find(
+        (blocktypeProperty) => blocktypeProperty.name === propertyName,
+      );
+
+      return !blocktypeProperty?.defaultValue;
```
georg-schwarz, line 53:
```suggestion
      return blocktypeProperty?.defaultValue === undefined;
```

libs/language-server/src/lib/validation/checks/block-definition.ts
```diff
@@ -73,13 +76,21 @@ function checkPipesOfBlock(
...
-      {
-        node: block,
-        property: 'name',
-      },
-    );
+    // The last block in a composite block is connected to the output, not another block
+    if (
+      !(
+        isCompositeBlocktypeDefinition(block.$container) &&
+        block.$container.blocks.at(-1)?.name === block.name
```
georg-schwarz, line 83: Consider an explanatory variable like `isLastBlockOfCompositeBlocktype`

libs/execution/src/lib/blocks/block-execution-util.ts
```diff
@@ -0,0 +1,107 @@
...
+
+export interface ExecutionOrderItem {
+  block: BlockDefinition;
+  value: IOTypeImplementation | null;
+}
+
+export async function executeBlocks(
+  executionContext: ExecutionContext,
+  executionOrder: ExecutionOrderItem[],
+  initialInputValue: IOTypeImplementation | undefined = undefined,
```
georg-schwarz, line 26: I don't entirely understand when this value is set. I assume when executing a composite block?

Maybe a good indicator to add JSDoc comment above explaining the parameter.

libs/execution/src/lib/blocks/block-execution-util.ts
```diff
@@ -0,0 +1,107 @@
...
+): Promise<R.Result<ExecutionOrderItem[]>> {
+  let isFirstblock = true;
```
georg-schwarz, line 28:
```suggestion
  let isFirstBlock = true;
```

libs/execution/src/lib/blocks/block-execution-util.ts
```diff
@@ -0,0 +1,107 @@
...
+    let inputValue =
+      parentData[0]?.value === undefined ? NONE : parentData[0]?.value;
+
+    if (
+      isFirstblock &&
+      inputValue === NONE &&
+      initialInputValue !== undefined
+    ) {
+      inputValue = initialInputValue;
+    }
```
georg-schwarz, line 44: In context of the comment above on the `initialInputValue` - maybe it makes sense to use an explanatory variable for the condition to express the code's intent more clearly.

libs/execution/src/lib/blocks/composite-block-executor.ts
```diff
@@ -0,0 +1,223 @@
...
+        );
+      }
+
+      this.removeVariablesFromContext(blockTypeReference.properties, context);
+
+      const pipeline = getPipeline(blockTypeReference);
+
+      if (R.isOk(executionResult)) {
+        // The last block always pipes into the output if it exists
+        const lastBlock = pipeline.blocks.at(-1);
```
georg-schwarz, line 98: Can be moved before the if-clause / line 94 can be moved down to have it in a cohesive place.

libs/execution/src/lib/blocks/composite-block-executor.ts
```diff
@@ -0,0 +1,223 @@
...
+      valueType: Valuetype,
+      block: BlockDefinition,
+      properties: BlocktypeProperty[],
+      evaluationContext: EvaluationContext,
+    ): InternalValueRepresentation | undefined {
+      const propertyFromBlock = block.body.properties.find(
+        (property) => property.name === name,
+      );
+
+      if (propertyFromBlock) {
```
georg-schwarz, line 170:
```suggestion
      if (propertyFromBlock !== undefined) {
```

georg-schwarz (approved):
Awesome! Looks quite intuitive after the refactor to me :)

I left a few very nitpicky comments. Feel free to ignore them as most of them are quite opinionated I guess ;-)

PR #422, opened 2023-08-15
[RFC] Multi-file Jayvee and Scoping

georg-schwarz:


rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
```diff
@@ -0,0 +1,116 @@
...
+**No-access paths:**
+- elements within a pipeline cannot be referenced by outside elements
+
+
+### Importing elements
+
+Only `libraries` and their elements can be imported into other files. Elements on the root level of a file or within a pipeline cannot be imported.
+
+```
+import './path/to/jv-file.jv'; // imports all libraries of the file
```
rhazn, line 90: I'd immediately talk of a `location` here by the way. We can by default map it to files but I feel it would be good to medium term have this more open to e.g. loading libraries from the web.

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
```diff
@@ -0,0 +1,116 @@
...
+  MyDomainLibrary called MyLibraryAlias
+}; // only imports the named library, access via qualified name using the alias
+import './path/to/jv-file.jv' use {
+  MyDomainLibrary.MyDomainSpecificValuetype called MyElementAlias
+}; // only imports the named element, access via the alias
+```
+
+References to these imported elements is by their qualified name (unless altered by an alias).
+
+## Drawbacks
```
rhazn, line 104: This is missing no way to version libraries. I'd consider this either:

- part of the location string
- explicit in the import, e.g. `from 'location' version 3.1.2 use { Name }` with the version string/no string being enforced semver. Could also force the library to define it's own version.

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  // ... possibly more elements
+}
+```
+
+Libraries are always "exported" and can be "imported" into other files.
+
+
+### Visibility of elements in a file
```
rhazn, line 59: I think we could consider, similarly to java, to force a match between files and language concepts. E.g. a .jv file can only contain one library or a pipeline, not both, not two libraries. That would create nice, clear 'library files' that only contain one library, nothing else and 'pipeline files' that contain a pipeline and optionally one of definitions of blocks and valuetypes. I'd consider that structure worth enforcing.

rhazn (changes requested):
Left some comments, I like the overall concept but I'd love your opinion mainly on the more enforced structure with one file per library.

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
georg-schwarz: While I generally agree that enforced structure might help users, I am reluctant to agree here with the Java example. It would lead to very bloated code, and in combination with the suggestion to remove block definitions from a `pipeline`, it would require jumping a lot of back and forth if you want to write reusable code.

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
georg-schwarz: I like the idea!

Enforcing it in the location string might be weird for local libraries I guess.
The explicit import syntax suggests there is a version per file. I think we should have rather a version per library (potentially multiple in a file). Thus, I suggest to use the syntax `from './path/to/location.jv' use { MyDomainLibrary version 1.2.3 };`

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
```diff
@@ -0,0 +1,116 @@
...
+import './path/to/jv-file.jv' use {
+  MyDomainLibrary.MyDomainSpecificValuetype called MyElementAlias
+}; // only imports the named element, access via the alias
+```
+
+References to these imported elements is by their qualified name (unless altered by an alias).
+
+## Drawbacks
+- Implicit knowledge required: elements in libraries are exported
+- Elements of a pipeline cannot be reused, leading to potentially more slim pipelines and a parallel library
```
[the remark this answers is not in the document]
georg-schwarz: We could do that, yes. You could even go further and only allow pipes in a pipeline for faster overview of the control flow. Not sure which version is best TBH.

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
```diff
@@ -0,0 +1,116 @@
...
+}; // only imports the named library, access via qualified name using the alias
+import './path/to/jv-file.jv' use {
+  MyDomainLibrary.MyDomainSpecificValuetype called MyElementAlias
+}; // only imports the named element, access via the alias
+```
+
+References to these imported elements is by their qualified name (unless altered by an alias).
+
+## Drawbacks
+- Implicit knowledge required: elements in libraries are exported
```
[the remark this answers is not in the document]
georg-schwarz: Ah that comes from the comment that libraries should not access blocks from the root block...
In that case, I'd make the elements in a library accessible unless marked with a keyword to keep them private (e.g., `hidden` or `private`).

A completely different approach would be to not introduce the concept `library` at all and add visibility keywords at root level (e.g., `share` or `public`).

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
rhazn: Hmm, I think that brings up the question if we want to import files and libraries the same. I like the explicit source/version etc import for libraries a lot, for files I think it is a bit of overkill. Maybe we could have a concept of including files and importing libraries?

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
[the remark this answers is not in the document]
georg-schwarz: I get your point. What's bugging me is having a concept `library` for sharing things and then not sharing per default but requiring another keyword. That seems like a design smell to me at first glance. But maybe I'm too nitpicky here :)

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
georg-schwarz: Right now, the RFC only covers importing libraries. Do we need to import a whole file at all?

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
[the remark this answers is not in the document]
rhazn: Ah yeah, fair enough if you see it in combination with it being in a library I can see the point. Then how about a middle ground approach initially where we just force a visibility statement (so no default)? Would go nicely with the explicit definition guideline ;).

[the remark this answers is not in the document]
georg-schwarz: Sure, that would be possible :)
I'd throw the alternative of removing the concept of a `library` again in the ring. Then we would have a keyword for elements on the root level to export them (e.g., `export`, `share`, `public`, ...)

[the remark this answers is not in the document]
rhazn: Hmm, I think the concept itself is nice. Libraries have more metadata (e.g., versions) than other code, plus I like the grouping of elements into one coherent library.

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
georg-schwarz: It makes sense, yeah. But that has a few implications, like having the visibility keyword at the root level as well, right? Or would you circumvent that?

Having two mechanisms in place allows us to be more specific, but also puts more load on the user. Especially if things are quite close together, that might be very confusing.
What if we remodel this RFC to only support local multi-file?
A future RFC would be to use `libraries` as an entrypoint to a versioned Jayvee model for reuse only for remote location import? That is also the case where we need versioning, for local import/export not so much.

rhazn: Hmm ya I think local file import first, then libraries would be good. Smaller and easier to execute on.

I like having visibility only in libraries, that way normal users don't need to use libraries (so have no visibility, no version, nothing, just their pipeline). And advanced users / library maintainers can deal with the overhead of maintaining a library if they actually want to do that.

georg-schwarz: Added it to the draft.

PR #423, opened 2023-08-21
Fixed constraint default property value retrieval

f3l1x98:
Fixed an issue that prevented the retrieval of default property values of constraint definitions

### DIFF
--- a/libs/execution/src/lib/execution-context.ts
+++ b/libs/execution/src/lib/execution-context.ts
@@ -9,12 +9,14 @@ import {
   ConstraintDefinition,
   EvaluationContext,
   InternalValueRepresentation,
+  MetaInformation,
   PipelineDefinition,
   PropertyAssignment,
   TransformDefinition,
   Valuetype,
   evaluatePropertyValue,
   getOrFailMetaInformation,
+  isConstraintDefinition,
   isExpressionConstraintDefinition,
   isPipelineDefinition,
   isPropertyBody,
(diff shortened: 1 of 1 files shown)


libs/execution/src/lib/execution-context.ts
```diff
@@ -129,9 +131,14 @@ export class ExecutionContext {
...
-    assert(isReference(currentNode.type));
 
-    const metaInf = getOrFailMetaInformation(currentNode.type);
+    let metaInf: MetaInformation;
+    if (isConstraintDefinition(currentNode)) {
+      metaInf = getOrFailMetaInformation(currentNode.type);
+    } else {
+      assert(isReference(currentNode.type));
+      metaInf = getOrFailMetaInformation(currentNode.type);
+    }
```
georg-schwarz, line 141: Can u do sth like 
```
if (isBlockDefinition(currentNode)) {
  assert(isReference(currentNode.type);
}
const metaInf = getOrFailMetaInformatin(currentNode.type);
```

Would be a little easier to read.

f3l1x98: That unfortunately does not work, because it would require a declaration of the following type: `getOrFailMetaInformation(type: Reference<BuiltinBlocktypeDefinition> | ConstraintTypeLiteral)`

The reason being that Typescript is unable to detect multiple declarations and decide which one to use.

PR #425, opened 2023-08-24
Add jayvee manifesto to dev docs

georg-schwarz:
I decided to add the Jayvee manifesto as the first section in the developer docs. Does it need any further elaboration?

### DIFF
--- a/apps/docs/docs/dev/01-intro.md
+++ b/apps/docs/docs/dev/01-intro.md
@@ -4,6 +4,20 @@ sidebar_position: 1
 
 # Introduction for Jayvee Developers
 
+## The Jayvee Manifesto
+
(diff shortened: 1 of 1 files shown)


apps/docs/docs/dev/01-intro.md
```diff
@@ -4,6 +4,20 @@ sidebar_position: 1
 
 # Introduction for Jayvee Developers
 
+## The Jayvee Manifesto
```
rhazn, line 7: I would make this it's own page and not just a section here. That way we can give some context when linking to it, e.g., that it is high level guidelines we want to follow when designing the language. Could then also be linked in all docs instead of just the dev ones.

rhazn (approved):
I like that it is online somewhere but I think the location could be refactored down the line 😅

georg-schwarz:
Closed in favor of #434

rhazn:
@georg-schwarz probably also can delete the branch, no?

PR #431, opened 2023-08-28
[WIP] TransformExecutor Tests

### LINKED ISSUE #415: [FEATURE] libs/execution tests
## User Story
1. As a {developer}
2. I want {to have tests to ensure that my new code does not break any existing code (without my knowledge)}
3. So that {I can publish better code}

## User Acceptance Criteria
- [x] Add constraint executor tests (#428)
- [ ] Add transform executor tests (if possible) (#431)

## Definitions of Done
- [ ] All tests are passing

f3l1x98:
Closes #415 

This PR adds tests for `TransformExecutor`.

### DIFF
--- a/libs/execution/src/lib/transforms/transform-executor.spec.ts
+++ b/libs/execution/src/lib/transforms/transform-executor.spec.ts
@@ -1,0 +1,289 @@
+// SPDX-FileCopyrightText: 2023 Friedrich-Alexander-Universitat Erlangen-Nurnberg
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+
+import assert = require('assert');
+
+import {
+  InternalValueRepresentation,
+  PrimitiveValuetypes,
+  TransformDefinition,
+  createJayveeServices,
+  useExtension,
+} from '@jvalue/jayvee-language-server';
+import {
+  ParseHelperOptions,
+  TestLangExtension,
+  expectNoParserAndLexerErrors,
+  parseHelper,
+  readJvTestAssetHelper,
+} from '@jvalue/jayvee-language-server/test';
+import { AstNode, AstNodeLocator, LangiumDocument } from 'langium';
+import { NodeFileSystem } from 'langium/node';
+
+import { getTestExecutionContext } from '../../../test/utils';
+import { Table, TableColumn } from '../types/io-types/table';
+
+import { PortDetails, TransformExecutor } from './transform-executor';
+
+describe('Validation of TransformExecutor', () => {
+  let parse: (
+    input: string,
+    options?: ParseHelperOptions,
+  ) => Promise<LangiumDocument<AstNode>>;
+
+  let locator: AstNodeLocator;
+
+  const readJvTestAsset = readJvTestAssetHelper(
+    __dirname,
+    '../../../test/assets/',
+  );
+
+  interface TableColumnDefinition {
+    columnName: string;
+    column: TableColumn;
+  }
+
+  function constructTable(
+    columns: TableColumnDefinition[],
+    numberOfRows: number,
+  ): Table {
+    const table = new Table(numberOfRows);
+    columns.forEach((col) => table.addColumn(col.columnName, col.column));
+    return table;
+  }
+
+  function getColumnsMap(
+    inputColumnNames: string[],
+    inputTable: Table,
+    transformInputDetailsList: PortDetails[],
+  ): Map<string, TableColumn> {
+    const variableToColumnMap: Map<string, TableColumn> = new Map();
+    for (let i = 0; i < inputColumnNames.length; ++i) {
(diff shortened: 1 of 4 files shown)


f3l1x98:
**Two things I have found:**
1. For the sake of creating a table as simple and fast as possible I added an constructor to the `Table` class. If I overlooked a reason as to why not do that please comment.
2. I wanted to add a test that would check if a row has been deleted because the expression could not be evaluated and I discovered that currently this only happens if there is no expression. The reason for that is that the `evaluateExpression` function throws assertion errors in case there is an issue which are not caught inside the `TransformExecutor#doExecuteTransform`. Is that intended, or should I add a try catch? (Because the logged error sounds like it should also handle the exceptions)

georg-schwarz (approved):

libs/execution/src/lib/transforms/transform-executor.spec.ts
```diff
@@ -134,18 +134,24 @@ describe('Validation of TransformExecutor', () => {
...
+      expect(result.resultingColumn.valuetype).toEqual(
+        PrimitiveValuetypes.Integer,
+      );
+      expect(result.resultingColumn.values).toHaveLength(1);
+      expect(result.resultingColumn.values).toEqual(
+        expect.arrayContaining([21]),
+      );
+    } catch (e) {
+      expect(e).toEqual(undefined);
+    }
```
georg-schwarz, line 154: Why do we need that try-catch here?

f3l1x98: Strictly speaking we don't just added it that all tests have the same format.

libs/execution/src/lib/transforms/transform-executor.spec.ts
```diff
@@ -0,0 +1,371 @@
...
+  ): Promise<{
+    resultingColumn: TableColumn<InternalValueRepresentation>;
+    rowsToDelete: number[];
+  }> {
+    const document = await parse(input, { validationChecks: 'all' });
+    expectNoParserAndLexerErrors(document);
+
+    const transform = locator.getAstNode<TransformDefinition>(
+      document.parseResult.value,
+      'transforms@0',
```
georg-schwarz, line 76: Could you pls iterate over all transforms? Makes the test setup less brittle.

libs/execution/src/lib/transforms/transform-executor.spec.ts
georg-schwarz: Then pls remove since it is kind of misleading

georg-schwarz:
> **Two things I have found:**
> 
> 1. For the sake of creating a table as simple and fast as possible I added an constructor to the `Table` class. If I overlooked a reason as to why not do that please comment.
> 2. I wanted to add a test that would check if a row has been deleted because the expression could not be evaluated and I discovered that currently this only happens if there is no expression. The reason for that is that the `evaluateExpression` function throws assertion errors in case there is an issue which are not caught inside the `TransformExecutor#doExecuteTransform`. Is that intended, or should I add a try catch? (Because the logged error sounds like it should also handle the exceptions)

Could you post a link to the file?
Assertions should never fail, they point to a bug.
If a user led to the mistake, there should be an exception IMO.

libs/execution/src/lib/transforms/transform-executor.spec.ts
f3l1x98: That would most likely not change anything because the execution of the other transforms will most likely crash due to wrong input / execution context

f3l1x98:
> > **Two things I have found:**
> > 
> > 1. For the sake of creating a table as simple and fast as possible I added an constructor to the `Table` class. If I overlooked a reason as to why not do that please comment.
> > 2. I wanted to add a test that would check if a row has been deleted because the expression could not be evaluated and I discovered that currently this only happens if there is no expression. The reason for that is that the `evaluateExpression` function throws assertion errors in case there is an issue which are not caught inside the `TransformExecutor#doExecuteTransform`. Is that intended, or should I add a try catch? (Because the logged error sounds like it should also handle the exceptions)
> 
> Could you post a link to the file? Assertions should never fail, they point to a bug. If a user led to the mistake, there should be an exception IMO.

I am talking about this commit (seems like I already added the try catch :sweat_smile: ): https://github.com/jvalue/jayvee/commit/28b5836b9c1363b4b02f04ca4332db8da80bc348

Because without this commit any exceptions thrown inside `evaluateExpression` will propagate through the whole stack up to the tests (or the system somewhere higher up in production).
Not sure why I wrote assertion errors instead of exceptions...

libs/execution/src/lib/transforms/transform-executor.spec.ts
f3l1x98: Discussed in person

georg-schwarz (approved):

Issue #430, opened 2023-08-28
Duplicate Logger: DefaultLogger and TestLogger


georg-schwarz:
The `DefaultLogger` is located in `interpreter-lib`, leading to in dependency cycle. We should move the `DefaultLogger` somewhere else and remove the `TestLogger` once usable.

_Originally posted by @georg-schwarz in https://github.com/jvalue/jayvee/pull/428#discussion_r1307025553_

f3l1x98:
@georg-schwarz I just thought of a potential use case for a dedicated test logger:
The execution implementations rarely "crash" but rather just log warnings etc, so I thought we could extend the test logger to keep a "cache" of logged messages. This would allow tests to check whether certain errors were logged etc.

We could also add an additional option to only write messages to the cache which could reduce logging output in the github actions.

PR #432, opened 2023-08-28
Test-logger extension

f3l1x98:
This adds some additional functionality to the `test-logger` as discussed https://github.com/jvalue/jayvee/issues/430#issuecomment-1695315374


georg-schwarz (changes requested):
Can you pls check if we can implement it by subclassing `DefaultLogger` instead to avoid duplicate logic?

f3l1x98:
> Can you pls check if we can implement it by subclassing `DefaultLogger` instead to avoid duplicate logic?

I intended to (that is why it is still in WIP, after all subclassing DefaultLogger will require some visibility changes in the logger)

georg-schwarz:
Ah wait, that will cause the dependency cycle again, right? Maybe we can move the default logger somewhere else?

f3l1x98:
> Ah wait, that will cause the dependency cycle again, right? Maybe we can move the default logger somewhere else?

Yes you are right.
I have moved the `default-logger.ts` into `execution/logging`.

georg-schwarz (commented):
LGTM. 

I have another use-case in mind: a program using Jayvee might want to extract the logs (instead of reading the std out).

Could you move it from the test to the regular library.

Another suggestion: store all logs in a common string array of a wrapper that lets you filter the different log levels. This way the order of log messages is preserved between the different log levels.

f3l1x98:
I could do that, but would do so in a separate PR given that this is another feature.

georg-schwarz (approved):

PR #437, opened 2023-09-01
Added CSVExtractor and GTFSExtractor to stdlib

rhazn:
This PR adds a CSVExtractor and a GTFSExtractor to the stdlib. However, they are not part of the autocomplete suggestions of the VSCode extension. The same is true for the Percentage valuetype though so I assume this is as expected for now, right @georg-schwarz ?

Part of #217

### DIFF
--- a/libs/language-server/src/stdlib/CSVExtractor.jv
+++ b/libs/language-server/src/stdlib/CSVExtractor.jv
@@ -9,8 +9,8 @@ composite blocktype CSVExtractor {
     property enclosing oftype text: '';
     property enclosingEscape oftype text: '';
 
-    input inputName oftype None;
-    output outputName oftype Sheet;
+    input inputPort oftype None;
+    output outputPort oftype Sheet;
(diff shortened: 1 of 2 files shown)


libs/language-server/src/stdlib/CSVExtractor.jv
```diff
@@ -0,0 +1,29 @@
...
+// SPDX-License-Identifier: AGPL-3.0-only
+
+// A CSVExtractor extracts a file from a given URL and interprets it as CSV
+composite blocktype CSVExtractor {
+    property url oftype text;
+    property delimiter oftype text: ',';
+    property enclosing oftype text: '';
+    property enclosingEscape oftype text: '';
+
+    input inputName oftype None;
```
georg-schwarz, line 12: That naming is a little weird, maybe use `inputPort` or so instead.

georg-schwarz (approved):

PR #438, opened 2023-09-01
Added documentation for composite blocks

rhazn:
localhost_3000_jayvee_docs_next_user_composite-blocks (1)

### DIFF
--- a/apps/docs/docs/user/composite-blocks.md
+++ b/apps/docs/docs/user/composite-blocks.md
@@ -1,0 +1,115 @@
+---
+sidebar_position: 4
+---
+
+# Composite Blocks
+
+Composite blocks are a way to create new blocktypes in Jayvee by combining the functionality of existing blocks and pipes. By relying on composite blocks instead of implementing more builtin blocks in a language interpreter Jayvee supports easy extension by users.
+
+Composite blocks define:
+- with the `property` keyword: properties with a name and [value type](./core-concepts.md#valuetypes), optionally a default value
+- with the `input` keyword: one input with a name and io type (that can be None)
+- with the `output` keyword: one output with a name and io type (that can be None)
+- one pipeline definition, starting from the input (using its name) and ending in the output (again using its name)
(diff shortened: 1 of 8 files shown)


apps/docs/docs/user/composite-blocks.md
```diff
@@ -0,0 +1,115 @@
+---
+sidebar_position: 4
+---
+
+# Composite Blocks
+
+Composite blocks are a way to create new blocktypes in Jayvee by combining the functionality of existing blocks and pipes. By relying on composite blocks instead of implementing more builtin blocks in a language interpreter Jayvee supports easy extension by users.
```
georg-schwarz, line 7:
```suggestion
Composite blocks are a way to create new blocktypes in Jayvee by combining the functionality of existing blocks and pipes. By relying on composite blocks instead of implementing more builtin blocks in a language interpreter, Jayvee supports easy extension by users.
```

apps/docs/docs/user/composite-blocks.md
```diff
@@ -0,0 +1,115 @@
...
+	block CarsCSVInterpreter oftype CSVInterpreter {
+		enclosing: '"';
+	}
+    // ... further block definitions
+}
+```
+
+### Refactoring using composite blocks
+
+The common usecase of downloading a CSV file using HTTP can be refactored into a composite block. Note that we define all properties of the builtin blocks that are used as properties of the new CSVExtractor blocktype (but add fallback values). If some internal configuration is always the same we could also not expose it as property of the new blocktype.
```
georg-schwarz, line 53:
```suggestion
The common use-case of downloading a CSV file using HTTP can be refactored into a composite block. Note that we define all properties of the builtin blocks that are used as properties of the new CSVExtractor blocktype (but add fallback values). If some internal configuration is always the same, we could also not expose it as a property of the new blocktype.
```

georg-schwarz (approved):
Love it!

PR #441, opened 2023-09-01
Add open source notices and update Sheetjs version

rhazn:
SheetJS community edition is the open-source version of the xslx library that was added.

They host their own code and we can not update to new versions with npm, they suggest vendoring in their install instructions:
https://docs.sheetjs.com/docs/getting-started/installation/nodejs/

They also require attribution in a specific form:
https://docs.sheetjs.com/docs/miscellany/license


georg-schwarz (approved):
LGTM. I like the download version since that can't introduce malicious code if their registry gets compromised (unless NPM where you cannot overwrite an existing version)

rhazn:
FYI also added https://github.com/jvalue/jayvee/issues/442 from this to automate complete license compliance.

georg-schwarz (approved):

PR #444, opened 2023-09-04
Document Constraint Types in Code

georg-schwarz:
This mechanism is similar to the current jv code generation to self-document blocktypes.

Exerpt:
image

I noticed some issue with Jayvee having no concept for INFINITY. Still, I'd vote to integrate this feature nevertheless.
image

### DIFF
--- a/libs/execution/src/lib/constraints/constraint-executor-registry.ts
+++ b/libs/execution/src/lib/constraints/constraint-executor-registry.ts
@@ -46,7 +46,8 @@ export function createConstraintExecutor(
   constraint: ConstraintDefinition,
 ): ConstraintExecutor {
   if (isTypedConstraintDefinition(constraint)) {
-    const constraintType = constraint.type.name;
+    const constraintType = constraint.type.ref?.name;
+    assert(constraintType !== undefined);
     const constraintExecutor = constraintExecutorRegistry.get(constraintType);
     assert(
       constraintExecutor !== undefined,
--- a/libs/execution/src/lib/constraints/constraint-executor-registry.ts
+++ b/libs/execution/src/lib/constraints/constraint-executor-registry.ts
@@ -47,7 +47,10 @@ export function createConstraintExecutor(
 ): ConstraintExecutor {
   if (isTypedConstraintDefinition(constraint)) {
     const constraintType = constraint.type.ref?.name;
-    assert(constraintType !== undefined);
+    assert(
+      constraintType !== undefined,
+      `Could not resolve reference to constraint type of ${constraint.name}`,
+    );
     const constraintExecutor = constraintExecutorRegistry.get(constraintType);
     assert(
       constraintExecutor !== undefined,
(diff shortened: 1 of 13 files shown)


libs/execution/src/lib/constraints/constraint-executor-registry.ts
```diff
@@ -46,7 +46,8 @@ export function createConstraintExecutor(
   constraint: ConstraintDefinition,
 ): ConstraintExecutor {
   if (isTypedConstraintDefinition(constraint)) {
-    const constraintType = constraint.type.name;
+    const constraintType = constraint.type.ref?.name;
+    assert(constraintType !== undefined);
```
rhazn, line 50: We can't keep doing asserts without error messages to handle errors. It has to have an error message, otherwise you just get a 'XYZ evaluated to a falsy value'.

libs/language-server/src/lib/builtin-library/stdlib.ts
```diff
@@ -118,6 +136,53 @@ function parseBuiltinValuetypeToJayvee(valuetype: PrimitiveValuetype): string {
...
+  Object.entries(metaInf.getPropertySpecifications()).forEach(
+    ([propName, propSpecification]) => {
+      const propDoc = propSpecification.docs?.description;
+      if (propDoc !== undefined) {
+        bodyLines.push(parseAsComment(propDoc, 1));
+      }
+      bodyLines.push(
+        `\tproperty ${propName} oftype ${propSpecification.type.getName()} ${
+          propSpecification.defaultValue !== undefined
+            ? `: ${internalValueToString(propSpecification.defaultValue)}`
```
rhazn, line 176: This should not parse to `Infinity` but to the actual number, even if it is a magic number. I am pretty sure it just uses Integer MAX_VALUE or something along those lines. Would get rid of the invalid jayvee code.

rhazn (changes requested):
If I understand this right this introduces builtin constraintypes without added logic and parses the ones from our Typescript code into a simulated stdlib jayvee code, right? If so, sounds like a good first step, mediumterm I think it would be good to go from actual jv code to our metainf. 

I disagree on the Infinity being okay though, I think we should never generate invalid jayvee code. I left a comment on how I think we can avoid that. Otherwise I am happy with this.

libs/language-server/src/lib/builtin-library/stdlib.ts
georg-schwarz: Addressed in [1d6cb98](https://github.com/jvalue/jayvee/pull/444/commits/1d6cb981023f4234a3cd67127d156ebc68ee06f3)

libs/execution/src/lib/constraints/constraint-executor-registry.ts
georg-schwarz: True, added in [6112703](https://github.com/jvalue/jayvee/pull/444/commits/611270315903f34152152fdbfaf27c2e4b6e3f55)

rhazn (approved):

PR #455, opened 2023-09-25
Test logger cache refactor

f3l1x98:
Refactored test logger cache as described in https://github.com/jvalue/jayvee/pull/432#pullrequestreview-1618660705


georg-schwarz (commented):
LGTM. Can you please move it out of the `test` project and call it `CachedLogger` or so? I think we will need this elsewhere as well.

georg-schwarz (approved):

PR #458, opened 2023-10-30
Tabular extension exec tests

### LINKED ISSUE #417: [FEATURE] tabular extension exec tests
## User Story
1. As a {developer}
2. I want {to have tests to ensure that my new code does not break any existing code (without my knowledge)}
3. So that {I can publish better code}

## User Acceptance Criteria
- [ ] Add tests to the exec part of the tabular extension

## Definitions of Done
- [ ] All tests are passing

f3l1x98:
Closes #417

This PR adds tests to all the block-executors of the tabular extension


f3l1x98:
I have found two things that need clarification:
1. https://github.com/jvalue/jayvee/blob/2240655e1c0c000635aaa05177cfb3eaab6eae43/libs/extensions/tabular/exec/src/lib/table-interpreter-executor.ts#L232-L234
Is this case-sensitive on purpose?
2. If the `TableInterpreterExecutor` is called for a Sheet with data (for example 8 rows of data) but with `columnDefinitions: [];` this will result into a table that satisfies the following: 
```
      // [...]
      const result = await parseAndExecuteExecutor(
        text,
        testWorkbook.getSheetByName('Sheet1') as R.Sheet,
      );
      // [...]
      const table = result.right;
      expect(table.getNumberOfColumns()).toEqual(0);
      expect(table.getNumberOfRows()).toEqual(8);
```
This is due to [`Table#addRow`](https://github.com/jvalue/jayvee/blob/f941c08ed57a9d7987a7d149af3e367192b85f23/libs/execution/src/lib/types/io-types/table.ts#L43) only verifying whether `rowLength === this.columns.size`, however that is always true in this case because [`TableInterpreterExecutor#constructAndValidateTableRow`](https://github.com/jvalue/jayvee/blob/2240655e1c0c000635aaa05177cfb3eaab6eae43/libs/extensions/tabular/exec/src/lib/table-interpreter-executor.ts#L155) returns an empty `TableRow` if there are no columns. Due to that `Table#addRow` will be executed without any issue and increment the row count, **however there is no data in the table**. 
**TLDR: Table indicates there are rows stored, even though there are none (there should not be any).**

georg-schwarz:
> 1. https://github.com/jvalue/jayvee/blob/2240655e1c0c000635aaa05177cfb3eaab6eae43/libs/extensions/tabular/exec/src/lib/table-interpreter-executor.ts#L232-L234
>    
>    Is this case-sensitive on purpose?

Yes, I believe so.

> 2. If the `TableInterpreterExecutor` is called for a Sheet with data (for example 8 rows of data) but with `columnDefinitions: [];` this will result into a table that satisfies the following:
> 
> ```
>       // [...]
>       const result = await parseAndExecuteExecutor(
>         text,
>         testWorkbook.getSheetByName('Sheet1') as R.Sheet,
>       );
>       // [...]
>       const table = result.right;
>       expect(table.getNumberOfColumns()).toEqual(0);
>       expect(table.getNumberOfRows()).toEqual(8);
> ```
> 
> This is due to [`Table#addRow`](https://github.com/jvalue/jayvee/blob/f941c08ed57a9d7987a7d149af3e367192b85f23/libs/execution/src/lib/types/io-types/table.ts#L43) only verifying whether `rowLength === this.columns.size`, however that is always true in this case because [`TableInterpreterExecutor#constructAndValidateTableRow`](https://github.com/jvalue/jayvee/blob/2240655e1c0c000635aaa05177cfb3eaab6eae43/libs/extensions/tabular/exec/src/lib/table-interpreter-executor.ts#L155) returns an empty `TableRow` if there are no columns. Due to that `Table#addRow` will be executed without any issue and increment the row count, **however there is no data in the table**. **TLDR: Table indicates there are rows stored, even though there are none (there should not be any).**

Good catch! I'd expect that only a row is added if there is data in it.
It might only make sense when you add all empty rows, then add content to the rows by adding columns. But I don't think we use this anywhere.

f3l1x98:
> Good catch! I'd expect that only a row is added if there is data in it.
It might only make sense when you add all empty rows, then add content to the rows by adding columns. But I don't think we use this anywhere.

We decided that this fix is ok, we just have to document this behavior somewhere.

georg-schwarz (approved):

PR #462, opened 2023-11-06
LoadTestExtensions utils function

f3l1x98:
Extracted the `initializeWorkspace` for test extension loading into a correspondingly named utils function for easier understanding

### DIFF
--- a/libs/interpreter-lib/src/parsing-util.spec.ts
+++ b/libs/interpreter-lib/src/parsing-util.spec.ts
@@ -8,11 +8,11 @@ import { CachedLogger } from '@jvalue/jayvee-execution';
 import {
   JayveeServices,
   createJayveeServices,
-  initializeWorkspace,
 } from '@jvalue/jayvee-language-server';
 import {
   ParseHelperOptions,
   expectNoParserAndLexerErrors,
+  loadTestExtensions,
   parseHelper,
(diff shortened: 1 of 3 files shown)


f3l1x98:
During this refactoring I also had the idea to move the definition of the test block types (e.g. `TestFileLoader` etc) into a builtin test lib. My idea was to load it the same way like [stdlib](https://github.com/jvalue/jayvee/tree/main/libs/language-server/src/stdlib) just only if run in test environment (option during `JayveeService` init or sth similar).

Not sure what you think about that idea @rhazn @georg-schwarz

georg-schwarz (approved):

rhazn:
> During this refactoring I also had the idea to move the definition of the test block types (e.g. `TestFileLoader` etc) into a builtin test lib. My idea was to load it the same way like [stdlib](https://github.com/jvalue/jayvee/tree/main/libs/language-server/src/stdlib) just only if run in test environment (option during `JayveeService` init or sth similar). Then we would not have to define the test extension for each test and the loading would be builtin and could be used should we implement jayvee pipeline tests for the user.

I like the idea of having a test lib of jv files that define built in block types that are only available during test runs. Sounds quite elegant to me. I am not that deep into the topic so I might miss something, but I like the idea.

PR #467, opened 2023-11-08
Remove process.exit commands

georg-schwarz:
In the `interpreter-lib`, there were still some `process.exit` commands. IMO, only surrounding applications should trigger a process exit.

The details of the `FAILURE` are logged / can be extracted by injecting a custom logger.

### DIFF
--- a/libs/interpreter-lib/src/parsing-util.spec.ts
+++ b/libs/interpreter-lib/src/parsing-util.spec.ts
@@ -27,8 +27,6 @@ describe('Validation of parsing-util', () => {
     options?: ParseHelperOptions,
   ) => Promise<LangiumDocument<AstNode>>;
 
-  let exitSpy: jest.SpyInstance;
-
   let services: JayveeServices;
 
   const logger = new CachedLogger(true, undefined, false);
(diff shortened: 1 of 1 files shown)


libs/interpreter-lib/src/interpreter.ts
```diff
@@ -82,16 +83,25 @@ export async function interpretModel(
             ', ',
           )}.`,
       );
-    process.exit(ExitCode.FAILURE);
+    return ExitCode.FAILURE;
```
georg-schwarz, line 86: I wonder if we should distinguish between different kinds of `FAILURES`, like `CONFIGURATION_FAILURE`, `RUNTIME_FAILURE`, `RUNTIME_ERROR`, and so on.

georg-schwarz:
I want to draft a release after integrating this. Should we do version `v0.2.0` to keep `v0.1.x` reserved for people using the current version and only supply bug fixes there? 
In theory, there should be no breaking changes - but I wouldn't guarantee for not having introduced bugs with the refactorings like #460.

libs/interpreter-lib/src/interpreter.ts
rhazn: Would be good, yeah. Note though, this is the exit code of the process so we can not freely define new ones (afaik). We definitely should have some form of more elaborate error output outside of the exit code though, like writing a logfile.

rhazn:
> I want to draft a release after integrating this. Should we do version `v0.2.0` to keep `v0.1.x` reserved for people using the current version and only supply bug fixes there? In theory, there should be no breaking changes - but I wouldn't guarantee for not having introduced bugs with the refactorings like #460.

I am not sure this doesn't change behavior. Was the cli-interpreter always also just returning exit code 1? Because otherwise, this is also a change in functionality that would mean 0.2.0 imho. But if there is no behavior change we should go with 0.1.1 and trust the tests I think?

Maybe in the future we should do some MADE only releases (e.g. release v.WS2324.MADE) so we can be more flexible with keeping a stable version for MADE and still release new versions for general users.

rhazn (approved):

PR #474, opened 2023-11-21
[RFC] Exports, Imports, and Libraries

georg-schwarz:
This PR is a reissue of #422.

## Major Changes
- Introduce two instead of one concept to enable multi-file Jayvee: exports and libraries for bundling of elements

### DIFF
--- a/rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
+++ b/rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
@@ -27,7 +27,7 @@ This feature will foster reuse of valuetypes, blocks, and other elements.
 Inherent to this feature is a concept of how scoping and naming is handled for nested structures.
 This RFC introduces two concepts:
 
-- File imports, and
+- Element usage from other files, and
 - Libraries
 
 ## Motivation
@@ -107,37 +107,37 @@ The **qualified name** is constructed by prepending container structures in this
 
 - elements within a pipeline cannot be referenced by outside elements
 
-**Imported elements are handled as if they were defined at the root level.**
+**Used elements of different files are handled as if they were defined at the root level.**
 
-### Importing elements
+### Using elements
 
-Only `publish`ed elements can be imported into other files.
+Only `publish`ed elements can be used in other files.
 
-#### Importing published elements of a file (within the same project)
+#### Using published elements of a file (within the same project)
 
 ```
-from './path/to/location.jv' use { MyDomainSpecificValuetype1 }; // only imports the defined elements from the file, access via qualified name as if it would be defined at the root level
-from './path/to/location.jv' use { MyDomainSpecificValuetype1 called Vt1} // only imports the defined elements from the file, access via qualified name using the alias
+from './path/to/location.jv' use { MyDomainSpecificValuetype1 }; // only use the defined elements from the file, access via qualified name as if it would be defined at the root level
+from './path/to/location.jv' use { MyDomainSpecificValuetype1 called Vt1} // only use the defined elements from the file, access via qualified name using the alias
 ```
 
-References to these imported elements is by their qualified name (unless altered by an alias).
+References to these used elements is by their qualified name (unless altered by an alias).
 
-#### Importing a library (from outside of the project)
+#### Using a library (from outside of the project)
 
-Each import explicitly defines the version of the imported library.
+Each `use` explicitly defines the version of the used library.
 On version mismatch, an error is raised.
-Libraries can only be imported as a whole.
+Libraries can only be used as a whole.
 
 ```
-from './path/to/location.jv' use { MyDomainLibrary version 1.2.3 }; // only imports the named library, access via qualified name
+from './path/to/location.jv' use { MyDomainLibrary version 1.2.3 }; // only use the named library, access via qualified name
 from './path/to/location.jv' use {
   MyDomainLibrary1 version 1.2.3,
   MyDomainLibrary1 version 1.2.3
-}; // only imports the named libraries, access via qualified name
-from './path/to/location.jv' use { MyDomainLibrary version 1.2.3 called MyLibraryAlias} // only imports the named library, access via qualified name using the alias
+}; // only use the named libraries, access via qualified name
+from './path/to/location.jv' use { MyDomainLibrary version 1.2.3 called MyLibraryAlias} // only use the named library, access via qualified name using the alias
 ```
 
-References to these imported elements is by their qualified name (unless altered by an alias).
+References to these used elements is by their qualified name (unless altered by an alias).
 
 ## Drawbacks
 
@@ -150,7 +150,7 @@ References to these imported elements is by their qualified name (unless altered
 ## Alternatives
 
 - "use" syntax without braces, etc., `from './path/to/file.jv' use MyDomainLibrary1, MyDomainLibrary2`
-- different syntax for importing files and libraries
+- different syntax for using files and libraries (e.g., `import`)
 - Rather call it `module` instead of `library`
 
 ## Possible Future Changes/Enhancements
--- a/rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
+++ b/rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
@@ -6,12 +6,13 @@ SPDX-License-Identifier: AGPL-3.0-only
 
 # RFC 0015: Multi-file Jayvee
 
-| | |
-|---|---|
-| Feature Tag | `multi-file` |
-| Status | `DRAFT` | <!-- Possible values: DRAFT, DISCUSSION, ACCEPTED, REJECTED -->
-| Responsible | `georg-schwarz` | <!-- TODO: assign yourself as main driver of this RFC -->
-<!-- 
+|             |                 |
+| ----------- | --------------- | --------------------------------------------------------------- |
+| Feature Tag | `multi-file`    |
+| Status      | `DRAFT`         | <!-- Possible values: DRAFT, DISCUSSION, ACCEPTED, REJECTED --> |
+| Responsible | `georg-schwarz` | <!-- TODO: assign yourself as main driver of this RFC -->       |
+
+<!--
   Status Overview:
   - DRAFT: The RFC is not ready for a review and currently under change. Feel free to already ask for feedback on the structure and contents at this stage.
   - DISCUSSION: The RFC is open for discussion. Usually, we open a PR to trigger discussions.
@@ -25,11 +26,11 @@ This RFC introduces the possibility of distributing a Jayvee program over multip
 This feature will foster reuse of valuetypes, blocks, and other elements.
 Inherent to this feature is a concept of how scoping and naming is handled for nested structures.
 
-
 ## Motivation
 
-Right now, Jayvee users can only pack their whole Jayvee model into one file. 
+Right now, Jayvee users can only pack their whole Jayvee model into one file.
 The challenge is two-fold:
+
 1. Larger projects will become unmaintainable quite quickly, as the elements cannot be organized into multiple files.
 2. Without distribution to multiple files, there is no possibility to reuse models of other projects.
 
@@ -40,9 +41,10 @@ For example, we might be enable to build libraries of valuetypes that can be reu
 ### Exporting elements
 
 For exporting elements, I propose introducing a new concept called `libraries` instead of explicitly modeling an import or export per element.
-A `library` can inhibit `Valuetype`s, `Block`s, `BlockType`s,  `Constraint`s, and `Transform`s.
+A `library` can inhibit `Valuetype`s, `Block`s, `BlockType`s, `Constraint`s, and `Transform`s.
 
 **Example library**
+
 ```
 library MyDomainLibrary {
   valuetype MyDomainSpecificValuetype {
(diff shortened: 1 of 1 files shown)


rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
```diff
@@ -0,0 +1,160 @@
...
+
+```
+export library MyDomainLibrary version 1.2.3 {
+  // definition of a new valuetype as part of the library
+  valuetype MyDomainSpecificValuetype1 {
+    // ... details of valuetype
+  }
+
+  // reference to an existing valuetype to make it part of the library
+  include MyDomainSpecificValuetype2;
```
georg-schwarz, line 77: I still dislike this one. Any ideas for a better syntax?

An alternative could be disallowing defining new elements in a library and only use it as an export list (potentially with the `called` keyword)

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
```diff
@@ -0,0 +1,160 @@
...
+
+SPDX-License-Identifier: AGPL-3.0-only
+-->
+
+# RFC 0015: Multi-file Jayvee
+
+|             |                 |
+| ----------- | --------------- | --------------------------------------------------------------- |
+| Feature Tag | `multi-file`    |
+| Status      | `DRAFT`         | <!-- Possible values: DRAFT, DISCUSSION, ACCEPTED, REJECTED --> |
```
rhazn, line 12:
```suggestion
| Status      | `DISCUSSION`         | <!-- Possible values: DRAFT, DISCUSSION, ACCEPTED, REJECTED --> |
```

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
```diff
@@ -0,0 +1,160 @@
...
+  - ACCEPTED: The RFC was accepted. Create issues to prepare implementation of the RFC.
+  - REJECTED: The RFC was rejected. If another revision emerges, switch to status DRAFT.
+-->
+
+## Summary
+
+This RFC introduces the possibility of distributing a Jayvee program over multiple files.
+This feature will foster reuse of valuetypes, blocks, and other elements.
+Inherent to this feature is a concept of how scoping and naming is handled for nested structures.
+We introduce two concepts in this RFC:
```
rhazn, line 28: Wording here is `we`, later on you use `I`. I would use neither, RFCs describe the state after they are accepted, not a suggestion. So this would be "Two concepts are introduced in this RFC:"

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
```diff
@@ -0,0 +1,160 @@
...
+1. Larger projects will become unmaintainable quite quickly, as the elements cannot be organized into multiple files.
+2. Without distribution to multiple files, there is no possibility to reuse models of other projects.
+
+For example, we might be enable to build libraries of valuetypes that can be reused across multiple projects instead of copying the code.
+
+## Explanation
+
+### Exporting elements for later import
+
+For exporting single elements, I propose to introduce a new keyword `export`.
```
rhazn, line 47: I think it would be actually nice to separate exporting and definition of elements. I'd suggest having an `export MyValueType;` and `export MyValueType as Name;` syntax. I'd also consider `publish` over `export`, I think `publish` sounds more natural with libraries (natural for non-programmers).

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
```diff
@@ -0,0 +1,160 @@
...
+The challenge is two-fold:
+
+1. Larger projects will become unmaintainable quite quickly, as the elements cannot be organized into multiple files.
+2. Without distribution to multiple files, there is no possibility to reuse models of other projects.
+
+For example, we might be enable to build libraries of valuetypes that can be reused across multiple projects instead of copying the code.
+
+## Explanation
+
+### Exporting elements for later import
```
rhazn, line 45: I am not sure if this is not overly complex. Why not be able to import local files in JV (just by `import 'filename.jv`) and treating that as replacing the import statement with the file content.

Export, import, renaming of exports etc all in libraries instead. Clear separation and design (imho).

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
rhazn: I don't understand why you need to include things here? Why not just require things to be defined in the lib?

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
```diff
@@ -0,0 +1,160 @@
...
+For bundling and exporting elements, I propose introducing a new concept called `libraries`.
+A `library` can inhibit `Valuetype`s, `Block`s, `BlockType`s, `Constraint`s, and `Transform`s.
+A library has to define a version in semver syntax.
+A library has to be exported.
+They serve as entry point to a collection of files (local or remote), similarly to JavaScript's `index.js` mechanism.
+
+**Example library**
+
+```
+export library MyDomainLibrary version 1.2.3 {
```
rhazn, line 70: Consider giving the library properties that make it more clear what the version is aka `breakingChange`, `compatibleChange`, `bugfixChange`? Can be mapped internally to semver but easier to understand for non-programmers?

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
```diff
@@ -0,0 +1,160 @@
...
+  // ... details
+}
+```
+
+### Bundling elements to a library for later import
+
+For bundling and exporting elements, I propose introducing a new concept called `libraries`.
+A `library` can inhibit `Valuetype`s, `Block`s, `BlockType`s, `Constraint`s, and `Transform`s.
+A library has to define a version in semver syntax.
+A library has to be exported.
```
rhazn, line 64: I'd consider not using the export keyword at all for a library and just going with the `publish` suggestion above. So library without any keyword and inside you have `publish` statements?

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
```diff
@@ -0,0 +1,160 @@
...
+They serve as entry point to a collection of files (local or remote), similarly to JavaScript's `index.js` mechanism.
```
rhazn, line 65: I don't think referencing another languages mechanism is very good in a RFC, especially not one as complex as JS bundling. The parts you mean should be in the RFC itself.

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
```diff
@@ -0,0 +1,160 @@
...
+**Imported elements are handled as if they were defined at the root level.**
+
+### Importing elements
+
+Only `export`ed elements can be imported into other files.
+
+#### Importing exported elements of a file
+
+```
+from './path/to/location.jv' use { MyDomainSpecificValuetype1 }; // only imports the defined elements from the file, access via qualified name as if it would be defined at the root level
```
rhazn, line 120: Imho `use { Name } from 'path to location'`, then the path can be a long URL and still look nice.

rhazn: Speaking of, this definitely needs clarity on what a valid path is. I'd consider HTTPs urls that lead to jv files (for a lib) valid as well as relative paths to file location (for importing a complete file or lib).

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
```diff
@@ -0,0 +1,160 @@
...
+
+- Two different sharing mechanisms (export keyword, library)
+- Elements of a pipeline cannot be reused, leading to potentially more slim pipelines and a parallel library
+- The elements of a library within a file always need the qualified name (alternative: allow access via sole name within file?)
+- We do not allow re-exporting (only by putting elements into a containing library)
+- Langium might not support this scoping mechanism out-of-the-box (more complex implementation)
+
+## Alternatives
+
+- "use" syntax without braces, etc., `from './path/to/file.jv' use MyDomainLibrary1, MyDomainLibrary2`
```
rhazn, line 153: I'd vote against.

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
```diff
@@ -0,0 +1,160 @@
...
+- different syntax for importing files and libraries
+- Rather call it `module` instead of `library`
```
rhazn, line 155: I'd vote against.

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Possible Future Changes/Enhancements
+
+- build out to use libraries of other projects via a package-manager mechanism, e.g., by using an URL as location of a "use" statement
```
rhazn, line 159: Ah, I'd have it in the RFC already but don't need to implement it at first.

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
joluj, line 70: I don't think that one should add the version in the code. Doing that will result in several problems:
- What happens if the same version but different code is uploaded?
- Resolving a version requires parsing every file in the library.

I think that using something like tagging (like in git) is better suited here.

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
joluj: +1, similar how deno handles the URL resolving maybe.

```
from './my-local.jv' use { Name }  // Local file
from 'https://jvalue.com/my-org/my-repo' use { Name }  // URL
from 'jv:my-org/my-repo' use { Name }  // Same as above, but one does not need a URL.
```

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
georg-schwarz: My thoughts were:
- if a library is the "entry point" to reusing a jv project, it might make sense to decouple it from the internal structure of the project
- For example, the user might define their valuetypes in a `my-valuetypes.jv`, but still wants to publish them with the library
- Otherwise, the library definition might become very large because it has to define everything that is accessible by other projects

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
georg-schwarz: `publish` it is!

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
georg-schwarz: Sure, can add that functionality. Proposed syntax: `from 'location' use *;`

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
georg-schwarz: How about we just add the concept of `library` here and leave the concrete features of it to another RFC? I think we should limit this RFC to only the things necessary and not further boat it. It already feels pretty complex.

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
georg-schwarz: Would you enforce that all elements within a library are published?

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
georg-schwarz: Suggestion: we define how importable locations look like (HTTP, file path) and leave the exact package-manager mechanism out of this PR (when to fetch remote libraries, where to cache them, etc.)

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
joluj: No, I'd enforce a `publish` keyword for the published elements. By default, nothing should be published.

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
georg-schwarz: Hmm, I agree that his requires some further thoughts. Would like to postpone it to a future RFC and remove the version mechanism in this RFC.

joluj: I agree with that. Just use the most up-to-date version in the current RFC.

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
georg-schwarz: Changed it to `publish <ElementName>;` as an alternative to definition within the library.

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
rhazn: Yeah I think that is a good plan. I also like the idea of having named versions, not semver but that can be a discussion in a follow up RFC then :D.

rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md
rhazn: If we do the "include element in library by using publish" then I agree, enforce publish in library.

georg-schwarz:
Close and reopen for a fresh discussion as changes became too many to properly discuss about them.

PR #478, opened 2023-12-01
[FEATURE] LocalFileExtractor block

### LINKED ISSUE #393: [FEATURE] Extract files from local file system
## User Story
1. As a {user}
2. I want {to extract files from my local file system}
3. So that {I can use them}

## User Acceptance Criteria
- [ ] A new block "LocalFileExtractor" is available, similar to the HttpExtractor that takes no input, creates a File output and has a property "path". It loads a file from the local file system, relative to the working directory of the interpreter
- [ ] A new block "LocalFolderExtractor" is available, same as LocalFileExtractor but output is a FileSystem

## Notes
- https://jvalue.github.io/jayvee/docs/dev/intro

## Definitions of Done
- [ ] All tests are passing

OmarFourati:
Related to #393 

Create logic and a new extractor block that would take care of local files

Closes #393


OmarFourati:
I have read the CLA Document and I hereby sign the CLA

OmarFourati:
@rhazn didn't manage to repair the pipeline.
It fails because of file-util not being found but it follows the exaxt logic of other files and should definitely be found. 
I need help there to be able to get review on this PR

libs/extensions/std/exec/src/local-file-extractor-executor.spec.ts
rhazn, line 1: There are tests missing for invalid file path properties that should also include edge cases. I'd suggest `../path-traversal-up.csv` and `./../../path-traversal-up.csv`.

libs/language-server/src/stdlib/builtin-blocktypes/LocalFileExtractor.jv
```diff
@@ -0,0 +1,22 @@
+// SPDX-FileCopyrightText: 2023 Friedrich-Alexander-Universitat Erlangen-Nurnberg
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+
+/**
+* Extracts a `File` from the local storage.
```
rhazn, line 6: Confusing naming since `localstorage` exists in browsers. I'd speak of file system.

```suggestion
* Extracts a `File` from the local file system.
```

libs/language-server/src/stdlib/builtin-blocktypes/LocalFileExtractor.jv
```diff
@@ -0,0 +1,22 @@
...
+* block CarsFileExtractor oftype LocalFileExtractor {
+*   filePath: "cars.csv";
+* }
+*/
+builtin blocktype LocalFileExtractor {
+	input default oftype None;
+	output default oftype File;
+	
+	/**
+	* The PATH to the file in the local storage to extract.
```
rhazn, line 18: Why `PATH`? There is a `PATH` env variable on basicaly any operating system so this is needlessly confusing. Just go with `path`. Same comment about `localstorage` as for the block description as well.

```suggestion
	* The path to the file in the local file system to extract.
```

rhazn (changes requested):
I think we are really close, well done!

We still have to:
- Ensure that no path traversal up can happen using regex instead of `.startsWith` (and test for it)
- Get the CI pipeline to work again

Regarding the CI pipeline I am honestly not sure what is happening there, the import looks identical to the `HttpExtractor` to me. But it has to work. Does it work locally for you?

rhazn (changes requested):
I think we are really close, well done!

We still have to:
- Ensure that no path traversal up can happen using regex instead of `.startsWith` (and test for it)
- Get the CI pipeline to work again

Regarding the CI pipeline I am honestly not sure what is happening there, the import looks identical to the `HttpExtractor` to me. But it has to work. Does it work locally for you?

libs/extensions/std/exec/src/local-file-extractor-executor.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+    nock.cleanAll();
+  });
+
+  it('should diagnose no error on valid local file path', async () => {
+    const text = readJvTestAsset('valid-local-file.jv');
+
+    const result = await parseAndExecuteExecutor(text);
+
+    expect(R.isErr(result)).toEqual(false);
+    if (R.isOk(result)) {
```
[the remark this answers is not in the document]
OmarFourati: image
Looks like it is done so in all the other tests.
Moreover not having that if throws an error the object result can't be called properly: 
```Property 'right' does not exist on type 'Result<BinaryFile>'.
  Property 'right' does not exist on type 'Left<ExecutionErrorDetails<AstNode>>'.ts(2339)```

File: libs/extensions/std/exec/src/local-file-extractor-executor.spec.ts [hunk 4]
```diff
@@ -0,0 +1,103 @@
...
+    }
+  });
+
+  it('should diagnose error on file not found', async () => {
+    const text = readJvTestAsset('invalid-file-not-found.jv');
+
+    const result = await parseAndExecuteExecutor(text);
+
+    expect(R.isErr(result)).toEqual(true);
+    if (R.isErr(result)) {
```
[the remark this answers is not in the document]
OmarFourati: same as before.

libs/extensions/std/exec/src/local-file-extractor-executor.spec.ts
OmarFourati: I sadly don't understand exactly what or how this should be implemented. With the regex implementation, I thought all the edge cases would through the error of the path traversal.
Need a hint there.
Thanks

libs/extensions/std/exec/src/local-file-extractor-executor.ts
```diff
@@ -0,0 +1,74 @@
...
+    input: None,
+    context: ExecutionContext,
+  ): Promise<R.Result<BinaryFile>> {
+    const filePath = context.getPropertyValue(
+      'filePath',
+      PrimitiveValuetypes.Text,
+    );
+
+    try {
+      const rawData = await fs.readFile(filePath);
```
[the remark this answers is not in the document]
OmarFourati: Done

OmarFourati:
> I think we are really close, well done!
> 
> We still have to:
> 
> * Ensure that no path traversal up can happen using regex instead of `.startsWith` (and test for it)
> * Get the CI pipeline to work again
> 
> Regarding the CI pipeline I am honestly not sure what is happening there, the import looks identical to the `HttpExtractor` to me. But it has to work. Does it work locally for you?

Yes it does. Locally all the tests and everything runs okey.
I don't understand neither. the util import doesn't show any error locally as well

libs/language-server/src/lib/validation/checks/blocktype-specific/property-assignment.ts
```diff
@@ -264,7 +264,9 @@ function checkLocalFileExtractorProperty(
   validationContext: ValidationContext,
 ) {
   if (propName === 'filePath') {
-    if (typeof propValue === 'string' && propValue.startsWith('../')) {
+    const validPathRegex = /^(?!.*\.\.\/).*$/;
```
rhazn, line 267: It would be much easier to test for an invalid path (which is anything that does include two dots I assume but please google a bit for if that makes sense). So I'd invert this check and check for invalid paths and if that is true, throw the error.

libs/extensions/std/exec/src/local-file-extractor-executor.spec.ts
rhazn: You should add tests that verify that you check the property correctly. This is done already in the `ZipArchiveInterpreter` we spoke about before. The two paths I mentioned in my original comment would be good examples to test, because they are both invalid and should throw errors related to invalid paths (not missing file).

I do not know what you mean by "With the regex implementation, I thought all the edge cases would through the error of the path traversal.", the test cases are to verify that is true, independently of the implementation.

rhazn (changes requested):

rhazn:
Closed in favor of https://github.com/jvalue/jayvee/pull/519 so we can work in a cleaner branch.

PR #487, opened 2023-12-15
Remove the SinglePipeDefinition Syntax

georg-schwarz:
Right now, we have two syntaxes to do the same thing.
I feel this leads to unnecessary complexity.
I suggest to remove the verbose `SinglePipeDefinition` syntax in favor of the `ChainedPipeDefinition` syntax using arrows.


rhazn (approved):

libs/language-server/src/lib/ast/wrappers/pipe-wrapper.ts
```diff
@@ -9,76 +9,42 @@ import { DiagnosticInfo } from 'langium';
...
-  SinglePipeDefinition,
-  isSinglePipeDefinition,
 } from '../generated/ast';
 
 import { AstNodeWrapper } from './ast-node-wrapper';
 
-export class PipeWrapper<N extends PipeDefinition = PipeDefinition>
-  implements AstNodeWrapper<N>
+export class PipeWrapper
+  implements AstNodeWrapper<PipeDefinition | BlocktypePipeline>
```
joluj, line 18: Why did you remove the generic? `astNode` can be narrowed down with the generic.

 ```ts
export class PipeWrapper<T extends PipeDefinition | BlocktypePipeline>
  implements AstNodeWrapper<T> { ... }
```

Current State:
```ts
declare const pipeline: BlocktypePipeline;
const wrapper = new PipeWrapper(pipeline, 29);
const node = wrapper.astNode; // <- typed as 'PipeDefinition | BlocktypePipeline' instead of 'BlocktypePipeline'
```

joluj (approved):

joluj:
@georg-schwarz
- I had to add a default type to prevent having to update the type in several places.
- I had to remove the improved typing of `getToDiagnostic` and `getFromDiagnostic` since the types went wild in `ValidationContext` and langium's `ValidationAcceptor`

georg-schwarz:
Awesome! Thx for cleaning up my mess @joluj :-)

Issue #488, opened 2023-12-18
[BUG] Scientific mathematical notations are being read as text


kreuz1995:
## Steps to reproduce
1. Scientific mathematical notations like 9E-2, which should actually mean 0.09, is being scanned as text in Jayvee (when tried to retrieve from a CSV file). Hence, when we keep the datatype of such a column decimal, a lot of the values are being filtered out.

Steps:
```
pipeline ThermoelectricMaterialsPipeline {

    block ThermoelectricMaterialsExtractor oftype HttpExtractor {
		url: "https://figshare.com/ndownloader/files/36554916";
	}

	block ZipArchiveInterpreter oftype ArchiveInterpreter {
		archiveType: "zip";
	}
	block MainThermoelectricMaterialsCSVPicker oftype FilePicker {
		path: "/TE_databases/CSV/main_tedb.csv";
	}
	block MainThermoelectricMaterialsTextFileInterpreter oftype TextFileInterpreter {
	}
	block MainThermoelectricMaterialsCSVInterpreter oftype CSVInterpreter {
		delimiter: ",";
	}
	block MainThermoelectricMaterialsTableInterpreter oftype TableInterpreter {
		header: true;
		columns: [
			"Name" oftype text,
            "Label" oftype text,
			"Editing" oftype text,
			"Model" oftype text,
			"Model_Type" oftype text,
			"Specifier" oftype text,
			"Value" oftype decimal,
			"Units" oftype text,
			"Temperature_Value" oftype text,
			"Temperature_Units" oftype text,
			"Value_Average" oftype text,
			"Temperature_Average" oftype text,
            "Pressure" oftype text,
			"Process" oftype text,
			"Direction_of_Measurement" oftype text,
			"Resolution" oftype text,
			"DOI" oftype text,
			"Title" oftype text,
			"Access_Type" oftype text,
			"Publisher" oftype text,
			"Publication_Year" oftype text,
			"Authors" oftype text,
			"Journal" oftype text,
		];
	}
	block MainThermoelectricMaterialsDatabaseLoader oftype SQLiteLoader {
		table: "MainThermoelectricMaterials";
		file: "./ThermoelectricMaterials.sqlite";
	}
	ZipArchiveInterpreter
		-> MainThermoelectricMaterialsCSVPicker
		-> MainThermoelectricMaterialsTextFileInterpreter
		-> MainThermoelectricMaterialsCSVInterpreter
		-> MainThermoelectricMaterialsTableInterpreter
[...]

PR #490, opened 2023-12-18
Adapted number regex to allow scientific notation

rhazn:
Allows parsing numbers in scientific notation, both integers like 1.904761905e+20 or decimals like 8.1e-5.

Closes https://github.com/jvalue/jayvee/issues/488

### DIFF
--- a/libs/execution/src/lib/types/valuetypes/internal-representation-parsing.ts
+++ b/libs/execution/src/lib/types/valuetypes/internal-representation-parsing.ts
@@ -11,7 +11,7 @@ import {
   ValuetypeVisitor,
 } from '@jvalue/jayvee-language-server';
 
-const NUMBER_REGEX = /^[+-]?([0-9]*[,\\.])?[0-9]+([eE][+\\-]?\d+)?$/;
+const NUMBER_REGEX = /^[+-]?([0-9]*[,.])?[0-9]+([eE][+-]?\d+)?$/;
 
 const TRUE_REGEX = /^true$/i;
 const FALSE_REGEX = /^false$/i;
--- a/libs/execution/src/lib/types/valuetypes/internal-representation-parsing.ts
+++ b/libs/execution/src/lib/types/valuetypes/internal-representation-parsing.ts
@@ -11,10 +11,7 @@ import {
   ValuetypeVisitor,
 } from '@jvalue/jayvee-language-server';
 
-const DECIMAL_COMMA_SEPARATOR_REGEX = /^[+-]?([0-9]*[,])?[0-9]+$/;
-const DECIMAL_DOT_SEPARATOR_REGEX = /^[+-]?([0-9]*[.])?[0-9]+$/;
-
-const INTEGER_REGEX = /^[+-]?[0-9]+$/;
+const NUMBER_REGEX = /^[+-]?([0-9]*[,\\.])?[0-9]+([eE][+\\-]?\d+)?$/;
 
 const TRUE_REGEX = /^true$/i;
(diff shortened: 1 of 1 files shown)


libs/execution/src/lib/types/valuetypes/internal-representation-parsing.ts
```diff
@@ -11,10 +11,7 @@ import {
   ValuetypeVisitor,
 } from '@jvalue/jayvee-language-server';
 
-const DECIMAL_COMMA_SEPARATOR_REGEX = /^[+-]?([0-9]*[,])?[0-9]+$/;
-const DECIMAL_DOT_SEPARATOR_REGEX = /^[+-]?([0-9]*[.])?[0-9]+$/;
-
-const INTEGER_REGEX = /^[+-]?[0-9]+$/;
+const NUMBER_REGEX = /^[+-]?([0-9]*[,\\.])?[0-9]+([eE][+\\-]?\d+)?$/;
```
georg-schwarz, line 14: I'm curious; what does the `[,\\.]` do?
I only know this syntax here to indicate a logical "or": `[,|.]`.
But I assume that is not what you want to express?

georg-schwarz (approved):
Should we document somewhere which kinds of notations we support?

libs/execution/src/lib/types/valuetypes/internal-representation-parsing.ts
```diff
@@ -47,24 +44,32 @@ class InternalRepresentationParserVisitor extends ValuetypeVisitor<
...
+    return Number.parseFloat(this.value.replace(',', '.'));
   }
 
   visitInteger(): number | undefined {
-    if (!INTEGER_REGEX.test(this.value)) {
+    /**
+     * Reuse decimal number parsing to capture valid scientific notation
+     * of integers like 5.3e3 = 5300. In contrast to decimal, if the final number
+     * is not a valid integer, returns undefined.
+     */
```
georg-schwarz, line 59: Love it!

libs/execution/src/lib/types/valuetypes/internal-representation-parsing.ts
rhazn: To my knowledge the logical or exists, so `,|\.` would be `,` or `.`. With the bracers, you define a group of characters that satisfy the regex, so `[,\.]` is "one of , or ." which is functionally identical to `,|\.`. The difference starts with things like `[0-9]` that you could write as `0|1|2|3|4|5|6|7|8|9`.

But since we have used `[+-]` before (and use `[eE]` later), I think sticking to these two character groups makes sense as well. Even though, seeing this I am not sure why I introduced the double escape in the second `[+-]`.

rhazn:
> Should we document somewhere which kinds of notations we support?

Probably, even though my hope is still that we will write parsers in Jayvee so they can be extended by users. I'd then move the notation documentation into their docs and autogenerate docs from there as we do for blocks. Otherwise I fear documentation will quickly get out of sync with implementation.

libs/execution/src/lib/types/valuetypes/internal-representation-parsing.ts
georg-schwarz: Wouldn't then that sub-expression mean "one of the following: , or  \ or any character" since the backslash is escaped instead of the fullstop?

rhazn: Yeah, I changed this again to not use the escaping character. I think (from the regex testers I used), the escape works as desired but it is just not needed in this case.

PR #491, opened 2023-12-19
Navigate Pipeline via Pipes instead of Blocks

georg-schwarz:
Until now, we have navigated over all blocks within a pipeline and connected them via pipes to get our execution order.
However, #485 will change that by allowing block definitions outside of pipelines, raising the necessity to iterate purely over the pipes with a pipeline.

This PR introduces a `PipelineWrapper` in the already existing of `AstNodeWrapper`s that inhibits the functionality to navigate over a pipeline and replaces the existing helper methods in the `model-util`.

The PR should not change the behavior of Jayvee besides:
- Pipelines with one block but no pipes were previously valid, which is not the case anymore (see adapted tests).

## TODOs:

- [x] make it work for `Pipelines`
- [x] make it work for `CompositeBlockTypes` (=> showcased by gtfs-static example)

### DIFF
--- a/example/gtfs-static.jv
+++ b/example/gtfs-static.jv
@@ -14,79 +14,78 @@ pipeline GtfsPipeline {
 	// 2. The origin for multiple pipe sequences is a zip
 	// file. Each csv file in this zip is further processed 
 	// by its own sequence of blocks and pipes.
-	GTFSSampleFeedExtractor -> ZipArchiveInterpreter;
 
-	ZipArchiveInterpreter
+	GTFSSampleFeedExtractor
 		-> AgencyFilePicker
 		-> AgencyTextFileInterpreter
 		-> AgencyCSVInterpreter
 		-> AgencyTableInterpreter
 		-> AgencyLoader;
 
-	ZipArchiveInterpreter	
+	GTFSSampleFeedExtractor	
 		-> CalendarDatesFilePicker
 		-> CalendarDatesTextFileInterpreter
 		-> CalendarDatesCSVInterpreter
 		-> CalendarDatesTableInterpreter
 		-> CalendarDatesLoader;
 
-	ZipArchiveInterpreter	
+	GTFSSampleFeedExtractor	
 		-> CalendarFilePicker
 		-> CalendarTextFileInterpreter
 		-> CalendarCSVInterpreter
 		-> CalendarTableInterpreter
 		-> CalendarLoader;
 
-	ZipArchiveInterpreter	
+	GTFSSampleFeedExtractor	
 		-> FareAttributesFilePicker
 		-> FareAttributesTextFileInterpreter
 		-> FareAttributesCSVInterpreter
 		-> FareAttributesTableInterpreter
 		-> FareAttributesLoader;
 
-	ZipArchiveInterpreter	
+	GTFSSampleFeedExtractor	
 		-> FareRulesFilePicker
 		-> FareRulesTextFileInterpreter
 		-> FareRulesCSVInterpreter
 		-> FareRulesTableInterpreter
 		-> FareRulesLoader;
 
-	ZipArchiveInterpreter	
+	GTFSSampleFeedExtractor	
 		-> FrequenciesFilePicker
 		-> FrequenciesTextFileInterpreter
 		-> FrequenciesCSVInterpreter
 		-> FrequenciesTableInterpreter
 		-> FrequenciesLoader;
 
-	ZipArchiveInterpreter			
+	GTFSSampleFeedExtractor			
 		-> RoutesFilePicker
 		-> RoutesTextFileInterpreter
 		-> RoutesCSVInterpreter 
 		-> RoutesTableInterpreter 
 		-> RoutesLoader;
 
-	ZipArchiveInterpreter			
+	GTFSSampleFeedExtractor			
 		-> ShapesFilePicker
 		-> ShapesTextFileInterpreter
 		-> ShapesCSVInterpreter 
 		-> ShapesTableInterpreter 
 		-> ShapesLoader;
 
-	ZipArchiveInterpreter	
+	GTFSSampleFeedExtractor	
 		-> StopTimesFilePicker
 		-> StopTimesTextFileInterpreter
 		-> StopTimesCSVInterpreter
 		-> StopTimesTableInterpreter 
 		-> StopTimesLoader;
 
-	ZipArchiveInterpreter
+	GTFSSampleFeedExtractor
 		-> StopsFilePicker 
 		-> StopsTextFileInterpreter
 		-> StopsCSVInterpreter 
 		-> StopsTableInterpreter 
 		-> StopsLoader;
 
-	ZipArchiveInterpreter	
+	GTFSSampleFeedExtractor	
 		-> TripsFilePicker 
 		-> TripsTextFileInterpreter
 		-> TripsCSVInterpreter 
@@ -94,12 +93,8 @@ pipeline GtfsPipeline {
 		-> TripsLoader;
 
 	// 3. As a first step, we download the zip file and interpret it.
-	block GTFSSampleFeedExtractor oftype HttpExtractor {
+	block GTFSSampleFeedExtractor oftype GTFSExtractor {
 		url: "https://developers.google.com/static/transit/gtfs/examples/sample-feed.zip";
-	}
-
-	block ZipArchiveInterpreter oftype ArchiveInterpreter {
-		archiveType: "zip";
 	}
 
 	// 4. Next, we pick several csv files (with the file extension ".txt") 
--- a/libs/execution/src/lib/blocks/block-execution-util.ts
+++ b/libs/execution/src/lib/blocks/block-execution-util.ts
@@ -54,7 +54,7 @@ export async function executeBlocks(
   for (const blockData of executionOrder) {
     const block = blockData.block;
     const parentData = pipelineWrapper
-      .getPredecessorBlocks(block)
+      .followIngoingPipes(block)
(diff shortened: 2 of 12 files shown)


example/gtfs-static.jv
```diff
@@ -14,94 +14,89 @@ pipeline GtfsPipeline {
...
+	GTFSSampleFeedExtractor	
 		-> TripsFilePicker 
 		-> TripsTextFileInterpreter
 		-> TripsCSVInterpreter 
 		-> TripsTableInterpreter 
 		-> TripsLoader;
 
 	// 3. As a first step, we download the zip file and interpret it.
-	block GTFSSampleFeedExtractor oftype HttpExtractor {
+	block GTFSSampleFeedExtractor oftype GTFSExtractor {
```
rhazn, line 96: I know this is not part of the PR but this seems very boring. We could enhance the `GTFSExtractor` in the std lib to also include picking a file and parsing it as CSV, maybe as a composite block inside a composite block.

The new block could be called `GTFSFileExtractor` and take an `url` and a `path` as property, then use the existing `GTFSExtractor` with the url and the `FilePicker` with the file, then parse that file as CSV file. This would reduce this example by 3 blocks and 3 lines for every file (so 33 less blocks and 33 less lines in the pipeline itself!).

libs/execution/src/lib/blocks/block-execution-util.ts
```diff
@@ -31,18 +34,31 @@ export interface ExecutionOrderItem {
  */
 export async function executeBlocks(
   executionContext: ExecutionContext,
-  executionOrder: ExecutionOrderItem[],
+  compositeBlockTypeDefinition:
```
rhazn, line 37: This is not well named if we allow a `PipelineDefinition` as well. How about `pipelineContainer`, `container` or similar?

libs/language-server/src/lib/ast/wrappers/pipeline-wrapper.ts
```diff
@@ -0,0 +1,83 @@
...
+
+export class PipelineWrapper
+  implements AstNodeWrapper<PipelineDefinition | CompositeBlocktypeDefinition>
+{
+  public readonly astNode: PipelineDefinition | CompositeBlocktypeDefinition;
+
+  allPipes: PipeWrapper[] = [];
+
+  constructor(
+    pipelineDefinition: PipelineDefinition | CompositeBlocktypeDefinition,
```
rhazn, line 22: If a parameter can be of multiple types, I wouldn't name it after only one. I'd suggest `container`, `pipelineContainer` or similar here as well.

libs/language-server/src/lib/ast/wrappers/pipeline-wrapper.ts
```diff
@@ -0,0 +1,83 @@
...
+    return withoutDuplicates;
+  }
+
+  getOutgoingPipes(blockDefinition: BlockDefinition): PipeWrapper[] {
+    return this.allPipes.filter((pipe) => {
+      return pipe.from === blockDefinition;
+    });
+  }
+
+  followOutgoingPipes(blockDefinition: BlockDefinition): BlockDefinition[] {
```
rhazn, line 70: I strongly feel we should follow terminology of directed acyclical graphs here.

This would mean for a block X:

- blocks reachable by following an outgoing pipe from X are called `children`
- blocks reachable by following an incoming pipe from X are called `parents`
- if you need to follow more than one pipe from X as described above the blocks are called `ancestors` or `descendants`

So for these functions it would mean renaming according to that:
- `followOutgoingPipes` would become `getChildren` (`getChildrenBlocks` works as well for me)
- `followIngoingPipes` would become `getParents` (`getParentBlocks`)

rhazn: This extends to variablenaming, e.g. `fromBlock` in line 48 would become `parentBlock`. I think this is a) much clearer and b) makes it much easier to read and understand what is going on here with the DAG image in the back of the mind.

libs/language-server/src/lib/ast/model-util.ts
```diff
@@ -20,101 +15,14 @@ import {
...
-    const blockType = new BlockTypeWrapper(block.type);
-
-    if (!blockType.hasInput()) {
-      result.push(block);
-    }
-  }
-  return result;
-}
-
-export function collectChildren(block: BlockDefinition): BlockDefinition[] {
```
rhazn, line 71: I like that we got rid of these util functions and moved to the wrapper concept we are using in other places already, well done :).

rhazn (changes requested):
I left some comments for barely relevant things outside of the scope of this PR (e.g. `GTFSExtractor` and I am happy if you ignore them.

I am however happy to die on the hill of using DAG naming (`parents`, `children` etc) so I'd ask for reverting the renaming to `followIngoing/OutgoingPipes`.

example/gtfs-static.jv
georg-schwarz: If you don't mind I'd leave it as it is for now. 

But your example looks interesting, also in the face of optimizations. In the best case, we would also reuse executed blocks multiple times. I'm not sure if that would be the case in the current implementation, so I'd leave it open for a follow-up issue.

libs/language-server/src/lib/ast/model-util.ts
georg-schwarz: Should I also relocate the function `getBlocksInTopologicalOrder`?

example/gtfs-static.jv
rhazn: Yeah, I agree on all counts, it's not in the scope of this PR anyway.

libs/language-server/src/lib/ast/model-util.ts
rhazn: I think that makes sense as well, yeah.

libs/execution/src/lib/blocks/block-execution-util.ts
georg-schwarz: Decided to go with `pipesContainer` in [021d04a](https://github.com/jvalue/jayvee/pull/491/commits/021d04a227e8ac9acc3f77fa14937e4ff8c037d9)

libs/language-server/src/lib/ast/wrappers/pipeline-wrapper.ts
georg-schwarz: Decided to go with `pipesContainer` in [c868cca](https://github.com/jvalue/jayvee/pull/491/commits/c868cca3414d172164ca61f4435e61d9f9cbd25e)

libs/language-server/src/lib/ast/wrappers/pipeline-wrapper.ts
georg-schwarz: Addressed in [78bba05](https://github.com/jvalue/jayvee/pull/491/commits/78bba058e40589972164b0a7725a64fc60ba153a)

libs/language-server/src/lib/ast/model-util.ts
georg-schwarz: Addressed in [2421058](https://github.com/jvalue/jayvee/pull/491/commits/2421058f33ee8ae19b802984baeed38a4d5ec244)

rhazn (approved):

libs/language-server/src/lib/validation/checks/block-definition.ts
```diff
@@ -39,14 +36,15 @@ function checkPipesOfBlock(
   }
   const blockType = new BlockTypeWrapper(block?.type);
 
+  const pipelineWrapper = new PipelineWrapper(block.$container);
```
georg-schwarz, line 39: Should check whether `canBeWrapped`

libs/language-server/src/lib/validation/checks/pipeline-definition.ts
```diff
@@ -22,7 +22,8 @@ function checkStartingBlocks(
   pipeline: PipelineDefinition,
   context: ValidationContext,
 ): void {
-  const startingBlocks = collectStartingBlocks(pipeline);
+  const pipelineWrapper = new PipelineWrapper(pipeline);
```
georg-schwarz, line 25: Should check whether `canBeWrapped`

libs/language-server/src/lib/ast/wrappers/pipeline-wrapper.ts
```diff
@@ -0,0 +1,141 @@
...
+  PipelineDefinition,
+} from '../generated/ast';
+
+import { AstNodeWrapper } from './ast-node-wrapper';
+import { PipeWrapper, createWrappersFromPipeChain } from './pipe-wrapper';
+
+export class PipelineWrapper
+  implements AstNodeWrapper<PipelineDefinition | CompositeBlocktypeDefinition>
+{
+  public readonly astNode: PipelineDefinition | CompositeBlocktypeDefinition;
```
joluj, line 19: I thought about transforming it into the following so that astNode is typed better. I'm not sure if we have a benefit from that though.

```suggestion
export class PipelineWrapper<
  T extends PipelineDefinition | CompositeBlocktypeDefinition,
> implements AstNodeWrapper<T>
{
  public readonly astNode: T;
```

joluj (approved):

libs/language-server/src/lib/ast/wrappers/pipeline-wrapper.ts
georg-schwarz: Awesome, seems to work now after removing the `SimplePipe` syntax in #487

joluj (approved):

PR #492, opened 2023-12-21
Add `--parse-only` flag

joluj:
This PR adds a `--parse-only` flag to the interpreter. Using this flag, the app just parses (and therefore validates) the given Jayvee model without executing it.

The process (when called with `--parse-only`) ends with `0` in case of valid model, with `1` otherwise.

### DIFF
--- a/apps/interpreter/src/index.ts
+++ b/apps/interpreter/src/index.ts
@@ -58,7 +58,7 @@ program
     undefined,
   )
   .option(
-    '--parse-only',
+    '-po, --parse-only',
     'Only parses the model without running it. Exits with 0 if the model is valid, with 1 otherwise.',
     false,
   )
--- a/apps/interpreter/src/index.ts
+++ b/apps/interpreter/src/index.ts
@@ -57,6 +57,11 @@ program
     `Sets the target blocks of the of block debug logging, separated by comma. If not given, all blocks are targeted.`,
     undefined,
   )
+  .option(
+    '--parse-only',
+    'Only parses the model without running it. Exits with 0 if the model is valid, with 1 otherwise.',
+    false,
+  )
   .description('Run a Jayvee file')
   .action(runAction);
 
--- a/apps/interpreter/src/parse-only.spec.ts
+++ b/apps/interpreter/src/parse-only.spec.ts
@@ -1,0 +1,44 @@
+import * as process from 'process';
+import { runAction } from './run-action';
+import { RunOptions } from '@jvalue/jayvee-interpreter-lib';
+import * as path from 'path';
+
+describe('Parse Only', () => {
+  const baseDir = path.resolve(__dirname, '../../../example/');
+
+  const defaultOptions: RunOptions = {
+    env: new Map<string, string>(),
+    debug: false,
+    debugGranularity: 'minimal',
+    debugTarget: undefined,
+    parseOnly: true,
+  };
+
+  beforeEach(() => {
+    jest.spyOn(process, 'exit').mockImplementation(() => {
+      throw new Error();
+    });
+  });
+
+  it('should exit with 0 on a valid option', async () => {
+    await expect(
+      runAction(path.resolve(baseDir, 'cars.jv'), {
+        ...defaultOptions,
+      }),
+    ).rejects.toBeDefined();
+
+    expect(process.exit).toBeCalledTimes(1);
+    expect(process.exit).toHaveBeenCalledWith(0);
+  });
+
+  it('should exit with 1 on error', async () => {
+    await expect(
+      runAction(path.resolve(baseDir, 'cars.jv'), {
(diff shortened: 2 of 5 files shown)


apps/interpreter/src/parse-only.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+      runAction(path.resolve(baseDir, 'cars.jv'), {
+        ...defaultOptions,
+      }),
+    ).rejects.toBeDefined();
+
+    expect(process.exit).toBeCalledTimes(1);
+    expect(process.exit).toHaveBeenCalledWith(0);
+  });
+
+  it('should exit with 1 on error', async () => {
```
rhazn, line 40: If I am not crazy this test runs the same file and also checks for a `0` exit code.

apps/interpreter/src/run-action.ts
```diff
@@ -23,6 +26,11 @@ export async function runAction(
       services,
       loggerFactory.createLogger(),
     );
+  if (options.parseOnly === true) {
+    const { model, services } = await parseModel(extractAstNodeFn, options);
+    const exitCode = model && services ? 0 : 1;
```
rhazn, line 31: At least in `parseModel` `model` and `services` are compared with `null`, not just falsy. Just for consistency I think checking for `null` here as well would make sense.

apps/interpreter/src/index.ts
```diff
@@ -57,6 +57,11 @@ program
     `Sets the target blocks of the of block debug logging, separated by comma. If not given, all blocks are targeted.`,
     undefined,
   )
+  .option(
+    '--parse-only',
```
rhazn, line 61: For consistenct also should have a shorthand, e.g. `-po`?

apps/interpreter/src/parse-only.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+  beforeEach(() => {
+    jest.spyOn(process, 'exit').mockImplementation(() => {
+      throw new Error();
+    });
+  });
+
+  it('should exit with 0 on a valid option', async () => {
+    await expect(
+      runAction(path.resolve(baseDir, 'cars.jv'), {
+        ...defaultOptions,
```
rhazn, line 32: These tests would be better if you explicitly define the `parseOnly` flag here instead of taking it from `defaultOptions`. That makes it easier to understand what is actually being tested here.

apps/interpreter/src/parse-only.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+      }),
+    ).rejects.toBeDefined();
+
+    expect(process.exit).toBeCalledTimes(1);
```
rhazn, line 36: There should be an assertion that tests that the model is only parsed, not interpreted (as that is the whole reason for this flag).

rhazn (changes requested):
Overall looks good. If I am not mistaken the tests are not testing what you want them to test though, that is what my request for changes is about.

rhazn (changes requested):
Overall looks good. If I am not mistaken the tests are not testing what you want them to test though, that is what my request for changes is about.

apps/interpreter/src/parse-only.spec.ts
```diff
@@ -2,34 +2,75 @@
...
+import {
+  RunOptions,
+  interpretModel,
+  interpretString,
+} from '@jvalue/jayvee-interpreter-lib';
 
 import { runAction } from './run-action';
 
+jest.mock('@jvalue/jayvee-interpreter-lib', () => {
+  const original: object = jest.requireActual('@jvalue/jayvee-interpreter-lib'); // Step 2.
```
rhazn, line 23: What is step 1? 🤔

apps/interpreter/src/parse-only.spec.ts
```diff
@@ -38,13 +79,25 @@ describe('Parse Only', () => {
...
   it('should exit with 1 on error', async () => {
+    const validModel = (await fs.readFile(pathToValidModel)).toString();
+
+    tempFile = path.resolve(
+      os.tmpdir(),
+      // E.g. "0.gn6v6ra9575" -> "gn6v6ra9575.jv"
+      Math.random().toString(36).substring(2) + '.jv',
+    );
+
+    // Write a partial valid model in that file
```
rhazn, line 90: It took me a few minutes to understand what this test is actually doing which I don't think is a good sign. Instead of 'partial valid model' I would make this explicit: 'Create an invalid model by truncating a previously valid model'. 

I think overall this is way overcomplicated though, why do you not just create an actually invalid model file and load it?

rhazn (approved):
Left a small style comment but overall happy with having this merged without further review 👍 .

georg-schwarz (approved):

Issue #497, opened 2024-01-09
[BUG] Composite Blocktype Validation does not account for different block ordering


georg-schwarz:
## Steps to reproduce
1. Add an unused block to an existing blocktype definition
2. Move the unused block up and down 

## Description
- Expected: The "unused" warning stays at the unused block
- Actual: The warning jumps over to other blocks

## Notes
- Issue is probably the validation function that assumes the correct order of blocks instead of navigating over the graph of blocks and pipes (see [here](https://github.com/jvalue/jayvee/blob/faecf34f8c2f71dd2fd22483bb916b31190eb58c/libs/language-server/src/lib/validation/checks/block-definition.ts#L81))
- In the same go, the validation logic should be moved to the container (pipeline, composite blocktype) to avoid "going upwards" in the AST as much as possible

TungstnBallon:
This seems to be fixed, perhaps by a44a73c?

[Screencast from 2024-05-17 12-06-38.webm](https://github.com/jvalue/jayvee/assets/59221538/e5a707a9-72c7-42f9-9fe2-f66356a9cc3f)

rhazn:
I agree, can't reproduce either.

PR #505, opened 2024-01-16
Simplify checking for incorrect extractor and loader blocks

georg-schwarz:
In the process of fixing #497, I'd like to simplify some validations.
I plan to refactor all validations in `validateBlockDefinition` because they should rather be located in `validatePipelineDefinition` and `validateCompositeBlockTypeDefinition`.
In this attempt, I noticed that we can simplify the validation whether a loader is connected via its output port, and whether an extractor is connected via its input port.

Improved the validation messages along the way..

## Before
Screenshot from 2024-01-16 10-32-03

## After
image

I think the error message is clear enough to cover this edge case as well.


joluj (approved):
The error message better but not very good imo. Why does it say SQLiteLoader if the block is a CarsLoader?

Maybe change it to
> The output type "None" of CarsLoader (as an SQLiteLoader) is incompatible with the input type "TextFile" of CarsCSVInterpreter (as a CSVInterpreter)

or simpler (preferred):
> The output type "None" of CarsLoader is incompatible with the input type "TextFile" of CarsCSVInterpreter

libs/language-server/src/lib/validation/checks/block-definition.ts
```diff
@@ -81,9 +61,18 @@ function checkPipesOfBlock(
       isCompositeBlocktypeDefinition(block.$container) &&
       block.$container.blocks.at(0)?.name === block.name;
 
+    const isExtractorBlock = !blockType.hasInput() && whatToCheck === 'input';
```
joluj, line 64: Naming is somewhat confusing since it may be an ExtractorBlock even if we don't check for it.

joluj (approved):
LGTM

libs/language-server/src/lib/validation/checks/block-definition.ts
georg-schwarz: Will be refactored in a future PR.

PR #506, opened 2024-01-16
Refactor block validation multiple pipe inputs

georg-schwarz:
- Changed logic to iterate over pipeline structure.
- Adapted some tests to be more precise

Validation output should have stayed roughly the same (except some minor edge cases)

### DIFF
--- a/libs/language-server/src/lib/validation/checks/block-definition.spec.ts
+++ b/libs/language-server/src/lib/validation/checks/block-definition.spec.ts
@@ -76,22 +76,6 @@ describe('Validation of BlockDefinition', () => {
     );
   });
 
-  it('should diagnose error on block as input for multiple pipes', async () => {
-    const text = readJvTestAsset(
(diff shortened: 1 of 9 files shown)


joluj (approved):

georg-schwarz:
A discussion emerged from this regarding #485 between @joluj and me. Probably of interest to @rhazn 

We currently don't model instances. This works until now as we prohibit scenarios where a block is reused by different inputting blocks (this validation was refactored in this PR). The question is if this is really desirable. If we allow it, we might run into ambiguous behavior.

Our takeaway was that we need to think ahead about how we want to model blocks with different ports. This seems to be the stage where we need to model instances latest. We did a whiteboard session on this and it feels really difficult to do without an instantiation logic where you can indicate whether a block is the same instance or if there are different instances of a block. Probably, this is also what caused confusion and is referenced by [this feedback](https://github.com/jvalue/jayvee/pull/485#issuecomment-1870320813).

PR #517, opened 2024-02-01
[FEATURE] Pipeline visualization

fabalex7:
Added an additional app "pipeline-visualization". 
This app can be used from the CLI by providing a .jv-file and a set of options. 
It produces the code for a visualization in mermaid-js format.
The output is logged onto the console and written to a separate .txt file. 
It can be included in .html sites or used with the mermaid [live editor] (https://mermaid.live/edit).


joluj:
This feature is postponed.

PR #521, opened 2024-02-08
Add a tertiary operator 'replace'

### LINKED ISSUE #482: [FEATURE] <title>
## User Story
1. As a {user}
2. I want {to remove brackets from the text datatype}
3. So that {to give it the correct datatype. So, if there is a column whose values are like "[integer]". it will be read as a text, just because of the brackets and cannot be used for any calculations or mathematical operations.}

## User Acceptance Criteria
- [ ] {acceptance criteria 1}: a transformation functions so that the brackets are removed.

## Examples

pipeline ThermoelectricMaterialsPipeline {

    block ThermoelectricMaterialsExtractor oftype HttpExtractor {
		url: "https://figshare.com/ndownloader/files/36554916";
	}

	block ZipArchiveInterpreter oftype ArchiveInterpreter {
		archiveType: "zip";
	}

	block InfThermoelectricMaterialsCSVPicker oftype FilePicker {
		path: "/TE_databases/CSV/inf_tedb.csv";
	}
	block InfThermoelectricMaterialsTextFileInterpreter oftype TextFileInterpreter {
	}
	block InfThermoelectricMaterialsCSVInterpreter oftype CSVInterpreter {
		delimiter: ",";

	}
	block InfThermoelectricMaterialsTableInterpreter oftype TableInterpreter {
		header: true;
		columns: [
			"Name" oftype text,
			"Temperature_Value" oftype text,
			"ZT" oftype text,
			"PF" oftype text,
			"S^2" oftype text,
			"s" oftype text,
			"k" oftype text,
			"Original_Counts" oftype text,
            "New_Counts" oftype text,
            "Inference" oftype text,
            "Editing" oftype text,
            "Pressure" oftype text,
            "Process" oftype text,
            "Label" oftype text,
            "Direction_of_Measurement" oftype text,
            "DOI" oftype text,
            "Publisher" oftype text,
            "Access_Type" oftype text,
            "Publication_Year" oftype text,
            "Authors" oftype text,
            "Journal" oftype text,
            "Comparison" oftype text,
		];
	}
	block InfThermoelectricMaterialsDatabaseLoader oftype SQLiteLoader {
		table: "InfThermoelectricMaterials";
		file: "./ThermoelectricMaterials.sqlite";
	}
	ThermoelectricMaterialsExtractor
		-> ZipArchiveInterpreter
		-> InfThermoelectricMaterialsCSVPicker
		-> InfThermoelectricMaterialsTextFileInterpreter
		-> InfThermoelectricMaterialsCSVInterpreter
		-> InfThermoelectricMaterialsTableInterpreter
		-> InfThermoelectricMaterialsDatabaseLoader;

In the data engineering pipeline, the temperature column has values like [200], which is read as a text by Jayvee. But in reality, it is just an integer inside the brackets, and due to the brackets not being removed, it is scanned as a text value, and cannot be used to set constraints or use for mathematical calculations

## Definitions of Done
- [ ] All tests are passing

rhazn:
Adds the `replace` operator.

This is:
- the first tertiary operator in Jayvee
- the first operator with a string return type

Closes #482.

### DIFF
--- a/libs/language-server/src/grammar/expression.langium
+++ b/libs/language-server/src/grammar/expression.langium
@@ -10,9 +10,12 @@ import './terminal'
 import './transform'
 
 Expression:
-  OrExpression;
+  ReplaceExpression;
 
 // The nesting of the following rules implies the precedence of the operators:
+
+ReplaceExpression infers Expression:
+  OrExpression ({infer TernaryExpression.first=current} operator='replace' second=OrExpression 'with' third=OrExpression)*;
 
 OrExpression infers Expression:
   AndExpression ({infer BinaryExpression.left=current} operator='or' right=AndExpression)*;
--- a/libs/language-server/src/lib/ast/expressions/evaluators/replace-operator-evaluator.ts
+++ b/libs/language-server/src/lib/ast/expressions/evaluators/replace-operator-evaluator.ts
@@ -1,0 +1,25 @@
+// SPDX-FileCopyrightText: 2023 Friedrich-Alexander-Universitat Erlangen-Nurnberg
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+
+// eslint-disable-next-line import/no-cycle
+import { DefaultTernaryOperatorEvaluator } from '../operator-evaluator';
+import { REGEXP_TYPEGUARD, STRING_TYPEGUARD } from '../typeguards';
+
+export class ReplaceOperatorEvaluator extends DefaultTernaryOperatorEvaluator<
+  string,
+  RegExp,
+  string,
+  string
+> {
+  constructor() {
+    super('replace', STRING_TYPEGUARD, REGEXP_TYPEGUARD, STRING_TYPEGUARD);
+  }
+  override doEvaluate(
+    firstValue: string,
+    secondValue: RegExp,
+    thirdValue: string,
+  ): string {
+    return firstValue.replace(secondValue, thirdValue);
+  }
+}
--- a/libs/language-server/src/lib/validation/validation-util.ts
+++ b/libs/language-server/src/lib/validation/validation-util.ts
@@ -15,6 +15,7 @@ import {
   isBinaryExpression,
   isExpressionLiteral,
   isFreeVariableLiteral,
+  isTernaryExpression,
   isUnaryExpression,
   isValueLiteral,
 } from '../ast';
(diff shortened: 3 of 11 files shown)


libs/language-server/src/grammar/expression.langium
```diff
@@ -10,10 +10,13 @@ import './terminal'
 import './transform'
 
 Expression:
-  OrExpression;
+  ReplaceExpression;
 
 // The nesting of the following rules implies the precedence of the operators:
 
+ReplaceExpression infers Expression:
+  OrExpression ({infer TernaryExpression.first=current} operator='replace' second=OrExpression 'with' third=OrExpression)*;
```
rhazn, line 18: I am not sure how comfortable I am that I understood this correctly. I think we should try to refactor this grammar to be more clearly readable (e.g. move all binary / ternary expressions into one type, have an easy and clear way to define order of execution instead of this weird setup of referring to the next lower expression).

In addition, I think the order I chose here means that the replace expression has the lowest priority in evaluation possible. Which I think makes sense (evaluate all other expressions first, then use their output for replacing)... but not sure.

libs/language-server/src/lib/ast/expressions/evaluators/replace-operator-evaluator.ts
```diff
@@ -0,0 +1,25 @@
...
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+
+// eslint-disable-next-line import/no-cycle
+import { DefaultTernaryOperatorEvaluator } from '../operator-evaluator';
+import { REGEXP_TYPEGUARD, STRING_TYPEGUARD } from '../typeguards';
+
+export class ReplaceOperatorEvaluator extends DefaultTernaryOperatorEvaluator<
+  string,
+  RegExp,
```
rhazn, line 11: This would be stronger as Regex OR string (so you could do `"Example" replace "ample" with "sample"` which is easier for non-programmers because you do not need to know regexes). I would like to add that in a follow up task if it is possible to do.

libs/language-server/src/lib/validation/validation-util.ts
```diff
@@ -135,6 +136,39 @@ function collectSubExpressionsWithoutFreeVariables(
       return [expression];
     }
     return [...leftSubExpressions, ...rightSubExpressions];
+  } else if (isTernaryExpression(expression)) {
+    const firstExpression = expression.first;
+    const secondExpression = expression.second;
+    const thirdExpression = expression.third;
```
rhazn, line 142: All logic after extracting the sub-expressions from the expression is the same as for binary expression... I think refactoring this into a shared logic would make sense and make this easier to read. Again, imho something for a follow up task.

rhazn:
I think we should build on this (or I should) with automated tests for operators, some additional operators (let me know some that you think make sense ;)) and a refactor of the code to be shorter and easier... but I would move that into a follow up task as well to keep this very minimal.

libs/language-server/src/lib/ast/expressions/evaluators/replace-operator-evaluator.ts
georg-schwarz: I think we would have to extend the `InternalValueRepresentation` to support type unions, then it should work.

georg-schwarz (approved):
> Imho out of scope (to keep the PR small):
> refactoring code to be easier to read or shorter, therefore I've largely copied the binary operator code and not unified them or changed the binary operator code

I agree, there is some code duplication. Seems not to be too excessive to me, though.

joluj (approved):

PR #528, opened 2024-02-12
Fixed publishing to npm failing because of peerDependencies

rhazn:
I am not sure why this did not fail earlier for me but this allows pack to run for all projects, using `npx nx run-many --target pack`. As this only changes the monaco editor used in the hub the impact should be small.


rhazn:
Also builds all projects as prod `npx nx run-many --target build --configuration=prod` successfully.

georg-schwarz (approved):

PR #531, opened 2024-02-16
RFC 0016 instances vs types

rhazn:
Continuing the discussion from #485.


rfc/0016-instances-vs-types/0016-instances-vs-types.md
```diff
@@ -0,0 +1,248 @@
...
+
+This implicit creation can lead to confusion between blocks and block instances. In addition, it makes it impossible to create more than one instance of a block, which will prevent the creation of more complex DAGs.
+
+This RFC introduces minimal syntax changes to allow the creation of block instances, introduces more code structure and clarifies wording.
+
+## Motivation
+It is impossible to create multiple instances of the same block, limiting how complex a DAG can be described.
+
+## Explanation
+The following explanations use blocks as an example but map directly to **Constraints** and **Valuetypes**. That is, constraints and valuetypes can also only be **defined** outside of pipelines and are valuetypes are **instantiated** inside of pipelines (constraints are only instantiated as part of valuetypes).
```
georg-schwarz, line 33: I think this text snippet belongs further down after the instantiation subject

rfc/0016-instances-vs-types/0016-instances-vs-types.md
```diff
@@ -0,0 +1,248 @@
...
+
+### Proposed changes
+#### 1. Clarify naming
+- Across M-levels, we have only a `instance of` relationship
+- On M1 level, only the `oftype` relationship exists between `block`s and `blocktype`s because it is modelled at M2 level
+- Instantiation of M2 level concepts in M1 is done with language keywords
+    - `composite blocktype <name>` / `builtin blocktype <name>`
+- We keep `block` as M1 level keyword to instantiate the M2 level `Block`
+- Users can describe M0 level entities in a Jayvee model
+    - Now by creating anonymous singleton instances of `block`s
```
georg-schwarz, line 168: Would leave that out here as we specify three different ways to do so further down.

rfc/0016-instances-vs-types/0016-instances-vs-types.md
```diff
@@ -0,0 +1,248 @@
...
+    "disp" oftype decimal,
+  ];
+}
+block CarsLoader oftype SQLiteLoader {
+  table: "Cars";
+  file: "./cars.sqlite";
+}
+```
+
+Composite blocks are a special case that use pipeline syntax to chain blocks. For the scope of this RFC, chaining pipeline syntax should **stay possible in pipelines and composite blocks** while **composite blocks are also be able to contain block definitions**. In a future enhancement (see Possible Future Changes/Enhancements), this should be split up with the introduction of packages (ref [RFC0015](../0015-multi-file-jayvee/0015-multi-file-jayvee.md)).
```
georg-schwarz, line 209:
```suggestion
Composite blocks are a special case that use a smiliar pipe syntax to chain blocks. For the scope of this RFC, chaining pipeline syntax should **stay possible in pipelines and composite blocks** while **composite blocks are also able to contain block definitions**. In a future enhancement (see Possible Future Changes/Enhancements), this should be split up with the introduction of packages (ref [RFC0015](../0015-multi-file-jayvee/0015-multi-file-jayvee.md)).
```

georg-schwarz (commented):
I left a few nitpicky comments. Overall, the RFC captures our intended changes very well I think.

rfc/0016-instances-vs-types/0016-instances-vs-types.md
```diff
@@ -0,0 +1,248 @@
...
+	}
+
+	block CarsLoader oftype SQLiteLoader {
+		table: "Cars";
+		file: "./cars.sqlite";
+	}
+}
+```
+
+##### As diagram including M2 level
```
joluj, line 83: I'd still strongly vote to use the conventions used in similar graphs. We use the names and semantics of `M0`, `M1`, and `M2` from the UML metamodel, so why do we decide to use the exact opposite semantic of the arrows?

 See [Metamodeling on Wikipedia](https://en.wikipedia.org/wiki/Metamodeling) and [an example that is very similar to our case](https://en.wikipedia.org/wiki/Metamodeling#/media/File:M0-m3.png).
Example

rfc/0016-instances-vs-types/0016-instances-vs-types.md
```diff
@@ -0,0 +1,248 @@
...
+CarsCSVExtractor -."oftype : OftypeRelationship".-> CSVExtractor
+CarsTableInterpreter -."oftype : OftypeRelationship".-> TableInterpreter
+CarsLoader -."oftype : OftypeRelationship".-> SQLiteLoader
+M0 ~~~ M1
+```
+
+### Proposed changes
+#### 1. Clarify naming
+- Across M-levels, we have only a `instance of` relationship
+- On M1 level, only the `oftype` relationship exists between `block`s and `blocktype`s because it is modelled at M2 level
```
joluj, line 163:
```suggestion
- The only other currently possible relationship is the `oftype` relationship between a `block` and a `blocktype` on M1 level. Other relationships are not possible since they are not modelled at M2 level.
```

rfc/0016-block-instantiation-syntax/0016-block-instantiation-syntax.md
```diff
@@ -0,0 +1,248 @@
...
+  D -> E;
+}
+```
+
+## Drawbacks
+- With additional enforced structure, users loose flexibility to define things where they want
+
+## Alternatives
+- Introduce explicit execution calls like `pipeline.execute();` or `instantiate <pipeline name>;`/`run <pipeline name>;` instead of implicitly executing pipelines
+  - Decided against because Jayvee is purely descriptive right now outside of expressions
```
joluj, line 242: We could go with calling it `declarative`. Or am I misinterpreting something here? https://en.wikipedia.org/wiki/Comparison_of_programming_paradigms
```suggestion
  - Decided against because Jayvee is purely declarative right now outside of expressions
```

rfc/0016-block-instantiation-syntax/0016-block-instantiation-syntax.md
```diff
@@ -0,0 +1,248 @@
...
+<!-- 
+  Status Overview:
+  - DRAFT: The RFC is not ready for a review and currently under change. Feel free to already ask for feedback on the structure and contents at this stage.
+  - DISCUSSION: The RFC is open for discussion. Usually, we open a PR to trigger discussions.
+  - ACCEPTED: The RFC was accepted. Create issues to prepare implementation of the RFC.
+  - REJECTED: The RFC was rejected. If another revision emerges, switch to status DRAFT.
+-->
+
+## Summary
+Following the [Design Principles](https://jvalue.github.io/jayvee/docs/dev/design-principles), Jayvee focuses on describing a goal state instead of how to get there. When creating pipelines, programmers use the pipe syntax `->` to connect previously defined `blocks`. Implicitely, the interpreter creates unnamed singleton instances of those blocks and executes them.
```
joluj, line 23:
```suggestion
Following our [Design Principles](https://jvalue.github.io/jayvee/docs/dev/design-principles), Jayvee focuses on describing a goal state instead of how to get there. When creating pipelines, programmers use the pipe syntax `->` to connect previously defined `blocks`. Implicitely, the interpreter creates unnamed singleton instances of those blocks and executes them.
```

joluj (approved):
Just two pedantic suggestions. I don't expect them to be changed, they are not worth a commit unless you change something else anyway.

PR #532, opened 2024-02-16
Replace block executor registry with dependency injection

georg-schwarz:
This PR solves an issue where the parallel execution of multiple Jayvee models led to side effects: an error by duplicate block executor registration. 

Injecting the list of executors wrapped by a `JayveeExecExtension` to the `ExecutionContext` instead of the registry should solve that issue.

Additionally, the executors were registered for model parsing. I think this is not necessary anymore as their structure now comes from `jv` files instead, so we should only use executors when interpreting the model.

The changes led to a weird trace of dependency cycles that I still struggle with. Let me know if you see an easy way to resolve this :)


libs/execution/src/lib/blocks/block-executor-registry.ts
```diff
@@ -6,10 +6,11 @@ import { strict as assert } from 'assert';
```
georg-schwarz, line 6: Probably should relocate that logic to `block-executor.ts`

apps/interpreter/src/examples-smoke-test.spec.ts
```diff
@@ -74,7 +73,6 @@ describe('jv example smoke tests', () => {
     sqliteLoaderMock.restore();
 
     // Clear registries
-    clearBlockExecutorRegistry();
     clearConstraintExecutorRegistry();
```
joluj, line 76: should we open an issue to remove `clearBlockExecutorRegistry()`?

joluj (approved):
The circular dependencies can lead to problems. However, I could find a better way.

I've found these circular dependencies. Some of these were not introduced by this PR.

```
$ npx madge --circular --extensions ts ./
Processed 72 files (422ms) (25 warnings)

✖ Found 10 circular dependencies!

1) src/lib/debugging/debug-configuration.ts > src/lib/execution-context.ts
2) src/lib/blocks/block-executor-class.ts > src/lib/blocks/block-executor.ts > src/lib/debugging/debug-configuration.ts > src/lib/execution-context.ts > src/lib/extension.ts
3) src/lib/types/io-types/filesystem-node-file-binary.ts > src/lib/types/io-types/io-type-implementation.ts
4) src/lib/types/io-types/io-type-implementation.ts > src/lib/types/io-types/filesystem-node-file-text.ts
5) src/lib/types/io-types/io-type-implementation.ts > src/lib/types/io-types/filesystem.ts
6) src/lib/types/io-types/io-type-implementation.ts > src/lib/types/io-types/none.ts
7) src/lib/types/io-types/io-type-implementation.ts > src/lib/types/io-types/sheet.ts
8) src/lib/types/io-types/io-type-implementation.ts > src/lib/types/io-types/table.ts
9) src/lib/types/io-types/io-type-implementation.ts > src/lib/types/io-types/workbook.ts
10) src/lib/blocks/block-execution-util.ts > src/lib/blocks/block-executor-registry.ts > src/lib/blocks/composite-block-executor.ts

```

apps/interpreter/src/examples-smoke-test.spec.ts
rhazn: Why not just remove it?

libs/execution/src/lib/blocks/block-executor-registry.ts
```diff
@@ -19,35 +20,35 @@ import {
...
 
-export const blockExecutorRegistry = new Registry<BlockExecutorClass>();
-
-export function registerBlockExecutor(executorClass: BlockExecutorClass) {
-  blockExecutorRegistry.register(executorClass.type, executorClass);
-}
-
-export function getRegisteredBlockExecutors(): BlockExecutorClass[] {
-  return [...blockExecutorRegistry.getAll()];
+export function getBlockExecutorClass(
```
rhazn, line 23: Wouldn't this be more natural if the method was defined on the execution extension? E.g. you'd use 

```
const execExtension = new StdExecExtension(); 
execExtension.getBlockExecutorClass("blockname");
```

instead of having to create an extension object and then calling a global function with that object as param as you do later in the stdextension test.

rhazn (approved):

apps/interpreter/src/examples-smoke-test.spec.ts
joluj: Ah I meant `clearConstraintExecutorRegistry`. I think we should do that but in another PR.

libs/execution/src/lib/blocks/block-executor-registry.ts
georg-schwarz: I wonder how that would play out when loading multiple extensions in the future... I think there is actually a need for a central component where you bring all the different extensions together, isn't it?

Or am I missing something?
Obviously, premature optimization is the source of all evil, so we can also remove it for now and introduce it later on (but in a separate PR?).

apps/interpreter/src/examples-smoke-test.spec.ts
georg-schwarz: Yeah, I think we should refactor that in a separate PR :+1:

libs/execution/src/lib/blocks/block-executor-registry.ts
georg-schwarz: Ah, I think I misunderstood your concern. Working on it!

apps/interpreter/src/examples-smoke-test.spec.ts
georg-schwarz: Decided to do it right away in this PR.

apps/docs/docs/dev/12-jayvee-testing.md
```diff
@@ -100,34 +100,23 @@ pipeline Pipeline {
...
-- `clearBlockExecutorRegistry` for clearing the registry containing all `BlockExecutor`s, and 
-- `clearConstraintExecutorRegistry` clearing the corresponding `ConstraintExecutor`s registry. 
-They are required in case the tested method initializes Jayvee itself (see [smoke test](#existing-tests-1)).
-
-[**test-logger.ts**](https://github.com/jvalue/jayvee/blob/dev/libs/execution/test/test-logger.ts):  
-This contains a subclass of the [`DefaultLogger`](https://github.com/jvalue/jayvee/blob/dev/libs/execution/src/lib/logging/default-logger.ts) used for tests which require a `Logger` implementation. The `TestLogger` contains the following tests functionality:
-- `getLogs`: retrieve the cached logs that the logger received.
-- `clearLogs`: clear the cached logs.
-
-[**block-executor-mocks.ts**](https://github.com/jvalue/jayvee/blob/dev/libs/execution/test/block-executor-mock.ts):
```
georg-schwarz, line 126: Outdated documentation

joluj (approved):

rhazn (approved):

Issue #454, opened 2023-09-19
[BUG] Fix wording of value types in documentation


dirkriehle:
## Problem

The documentation uses inconsistent forms for the concept of value type. There is "value type", "Valuetype", "valuetype", etc.

## Solution

Settle on the use of "value type" and capitalize the V in "value" only if English grammar requires it.

In implementation or code, turn the space between "value" and "type" into camel case, i.e. "ValueType".

dirkriehle:
Comments welcome; I can try my hands at fixing it in the user documentation.

rhazn:
I agree, would be happy with a PR.

georg-schwarz:
**Hint:** the docs are located as markdown files here: https://github.com/jvalue/jayvee/tree/main/apps%2Fdocs%2Fdocs

dirkriehle:
Sure, will do.

PR #539, opened 2024-04-05
Use camel case for value type (valueType) internally

georg-schwarz:
First steps towards #454.

This PR introduces consistent naming of classes and variables; only in code.

Does **not** fix all occurrences in the docs. We should update the docs consistently, e.g., using `block types` instead of `BlockTypes` as well. Otherwise, the docs make an inconsistent impression. I touched the docs at a few places but left out the very obvious ones where a deviating naming scheme jumps to the eye immediately.

I suggest refactoring the docs before making the code used for `blockType`, etc, consistent.


joluj (commented):
Should we update RFCs 7, 8, 11, 15, 16?

georg-schwarz:
> Should we update RFCs 7, 8, 11, 15, 16?

Probably not. Reverted the changes on them.

joluj (approved):

PR #544, opened 2024-04-12
fix(FilePicker): path can now start with a dot

### LINKED ISSUE #381: [BUG] File picker should not crash for paths starting with a dot
## Steps to reproduce
1. Use a file `FilePicker` block with a path for an existing file, starting with a dot: `./data.csv`

## Description
- Expected: The `FilePicker` selects the file
- Actual: `error: An unknown error occurred: AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:
  (0, assert_1.strict)(file instanceof jayvee_execution_1.BinaryFile)`

## Notes
- using the path without a dot (`/data.csv`) works
- Reported in studon: https://www.studon.fau.de/studon/goto.php?target=frm_5003028_338627_989410&client_id=StudOn

TungstnBallon:
paths inside `FilePicker` can now start with a dot

```
block DataPicker oftype FilePicker {
	path: "./data.csv";
}
```

closes #381

### DIFF
--- a/libs/execution/src/lib/types/io-types/filesystem-inmemory.ts
+++ b/libs/execution/src/lib/types/io-types/filesystem-inmemory.ts
@@ -57,12 +57,12 @@ export class InMemoryFileSystem implements FileSystem {
   }
 
   private processPath(path: string): string[] | null {
-    if (!path.startsWith(InMemoryFileSystem.getPathSeparator())) {
+    const [head, ...tail] = path.split(InMemoryFileSystem.getPathSeparator());
+    if (!(head === '' || head === InMemoryFileSystem.CURRENT_DIR)) {
       return null;
     }
-    const parts = path
-      .split(InMemoryFileSystem.getPathSeparator())
-      .filter((p) => p !== ''); // Process paths like "folder1//folder1" to "folder1/folder2"
+
+    const parts = tail.filter((p) => p !== ''); // Process paths like "folder1//folder1" to "folder1/folder2"
     const processedParts: string[] = [];
     for (const part of parts) {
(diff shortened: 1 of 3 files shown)


rhazn (approved):
If possible, a great practice when fixing bugs is:
- Create a unit test that fails with the current bug (here, create a test for the `FilePicker` blocktype that fails with a leading dot path)
- Fix the bug

This workflow ensures that the first time a bug is reported it is the last time a human encounters it.

In this case, I think we have unit tests for block types so I'd ask you to add one for this situation. That can be in a separate PR though, after merging this one. You can use "references #381" in the description of that PR to make sure they are nicely linked

rhazn:
(In case you'd like me to review this again, please re-request a review on the top right @TungstnBallon :)).

georg-schwarz:
Hey @TungstnBallon,

Sorry that this happened the second time now, but there has been a refactoring again :sweat_smile: 
#545 renamed all occurrences of "blocktype" to "blockType", it probably affects some of your imports.

Let me know if I can help updating the branch ;-)

TungstnBallon:
is there an advantage to putting the tests in a separate pull request? @rhazn

rhazn (approved):
No advantage to do it in a separate PR in general (in fact, changes and their tests should be in one PR). But sometimes it is better to get a smaller PR merged as soon as possible and move related work to future PRs. For example to avoid what happened here with Georg making changes that conflicted with yours in the meantime.

I wasn't sure if you'd need more time to figure out how testing in Jayvee works, but it looks like it went fast so great job 👍.

You can merge this, I'd squash and merge it so it becomes one consistent commit though.

Issue #427, opened 2023-08-24
[FEATURE] Update Langium to 2.0


rhazn:
## UAC
- [ ] Langium is updated to version 2 https://github.com/eclipse-langium/langium/blob/main/packages/langium/CHANGELOG.md#v200-aug-2023

## Notes
- Consider if it is easier to do 1.3 first https://github.com/jvalue/jayvee/issues/427

## Definitions of Done
- [ ] A PR has been opened and accepted
- [ ] All user acceptance criteria are met
- [ ] All tests are passing

georg-schwarz:
Thru some experiments, I found the following links very helpful for a future migration
- you can create commonjs and esm outputs in parallel for a potentially gradual migration to esm: https://www.sensedeep.com/blog/posts/2021/how-to-create-single-source-npm-module.html
- there is a cli tool that does most of the import rewriting for you: https://github.com/bennycode/ts2esm

PR #546, opened 2024-04-15
Encourage Type Imports

georg-schwarz:
On our strive towards migrating to Langium 2 (#427), we approached import cycles as a roadblock that we have to get rid of first, as ESM seems to struggle with them while CommonJS was able to handle them.

In this process, I noticed a fair amount of unnecessary obfuscation in these import cycles: when we only import types instead of values, the import cycle detection should ignore the occurrence. Right now, it cannot do that, as we only rarely use type imports, we import as value most of the time. Thus, the import cycles contain many unnecessary elements that make wrapping your head around the real problem more difficult than it has to be.

By adding the linter rule [@typescript-eslint/consistent-type-imports](https://typescript-eslint.io/rules/consistent-type-imports/), the import cycles are reduced to their minimum of elements. Fortunately, `eslint` was able to auto-fix almost all occurrences.

All changes directly follow the change in commit 8a9067bc033d9fc4c12b669042fb1d20c0d4bbb5 where the new linter rule is added as a `warning`.


georg-schwarz:
No need to review all changes, they came automatically with the linter auto-fix. But I'd like to hear your thoughts on whether to add or not add the linter rule ;-)

joluj:
I like the overall idea, however, I'd set the configuration `fixStyle` to `inline-type-imports`.
This should result is cleaner/fewer imports.

```ts
// Before this PR
import { AbstractBlockExecutor, BlockExecutor } from './block-executor';

// Current setting (default / `separate-type-imports`)
import type { BlockExecutor } from './block-executor';
import { AbstractBlockExecutor } from './block-executor';

// My suggestion (`inline-type-imports`)
import { AbstractBlockExecutor, type BlockExecutor } from './block-executor';
```

This is supported since typescript 4.5, we currently use ts 5.0 so there should be no problem.

rhazn (approved):
Works for me and I agree with Johannes.

PR #547, opened 2024-04-15
Remove dependency cycle caused be CellRangeWrapper being an internal value representation

georg-schwarz:
Removes a dependency cycle from our code base by using the `WrapperFactoryProvider` construct introduced in #536.

The remaining dependency cycles revolve around our design of the `ValueType` and `wrapper` constructs.

### DIFF
--- a/libs/language-server/src/lib/ast/expressions/index.ts
+++ b/libs/language-server/src/lib/ast/expressions/index.ts
@@ -5,9 +5,9 @@
 export * from './evaluation-strategy';
 
 export * from './evaluation-context';
-// eslint-disable-next-line import/no-cycle
 export * from './internal-value-representation';
 export * from './operator-registry';
+// eslint-disable-next-line import/no-cycle
 export * from './type-inference';
 export * from './typeguards';
 export * from './evaluate-expression';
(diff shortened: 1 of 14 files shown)


libs/language-server/src/lib/ast/expressions/internal-value-representation.ts
```diff
@@ -4,16 +4,17 @@
...
   type TransformDefinition,
   type ValuetypeAssignment,
   isBlockTypeProperty,
+  isCellRangeLiteral,
   isConstraintDefinition,
   isTransformDefinition,
   isValuetypeAssignment,
 } from '../generated/ast';
-// eslint-disable-next-line import/no-cycle
-import { type CellRangeWrapper, isCellRangeWrapper } from '../wrappers';
```
georg-schwarz, line 16: This import cycle is resolved in this PR.

libs/language-server/src/lib/ast/expressions/index.ts
```diff
@@ -5,9 +5,9 @@
 export * from './evaluation-strategy';
 
 export * from './evaluation-context';
-// eslint-disable-next-line import/no-cycle
 export * from './internal-value-representation';
 export * from './operator-registry';
+// eslint-disable-next-line import/no-cycle
```
georg-schwarz, line 10: This import cycle should have been masked by the removed one

rhazn (approved):

PR #553, opened 2024-04-18
Remove Import Cycle in type-inference.ts

georg-schwarz:
Removes the last dependency cycles indicated by the eslint `no-cycle` rule :partying_face:

### DIFF
--- a/libs/language-server/src/lib/ast/wrappers/value-type/index.ts
+++ b/libs/language-server/src/lib/ast/wrappers/value-type/index.ts
@@ -2,6 +2,8 @@
 //
 // SPDX-License-Identifier: AGPL-3.0-only
 
+export * from './util';
+
 /*
  * Note: Only export types if possible to enforce usage of WrapperFactory outside this directory.
  * This allows us to avoid dependency cycles between the language server and interpreter.
--- a/libs/language-server/src/lib/ast/wrappers/value-type/index.ts
+++ b/libs/language-server/src/lib/ast/wrappers/value-type/index.ts
@@ -2,8 +2,6 @@
 //
 // SPDX-License-Identifier: AGPL-3.0-only
 
-export * from './util';
-
 /*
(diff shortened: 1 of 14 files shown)


libs/language-server/src/lib/ast/wrappers/value-type/util.ts
```diff
@@ -0,0 +1,91 @@
...
+    const current = types[i - 1]!;
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const afterCurrent = types[i]!;
+    if (!current.equals(afterCurrent)) {
+      return false;
+    }
+  }
+
+  return true;
+}
```
georg-schwarz, line 91: Functions could be moved to the `ValueTypeProvider` introduced in #549. Opinions?

rhazn: I have no strong opinion but I see no harm to keep it as util function and not add it to an (only semi-related as far as I can see) class.

libs/language-server/src/lib/ast/wrappers/value-type/index.ts
```diff
@@ -2,6 +2,8 @@
 //
 // SPDX-License-Identifier: AGPL-3.0-only
 
+export * from './util';
```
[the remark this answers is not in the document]
georg-schwarz: I think there is a slight difference in those two exports that justify one being affected by the comment and one not.

In line 5, we just blindly export everything in the file (no matter if type or not) - it's not related to the WrapperFactory.
In line 14, we do care about type exports (so I added a comment that it is handled by the index.ts file) - it's related to the WrapperFactory.

It might be a code smell TBH.. I moved the value type utils one level up to the `wrappers/utils` directory ([9ae2670](https://github.com/jvalue/jayvee/pull/553/commits/9ae26707622f359b862f8440e13a37802a21399a)). Hope that makes it clearer :)

PR #556, opened 2024-04-19
Upgrade to Langium 2 and ESM

georg-schwarz:
Also replaces `jest` with `vitest` as jest has some troubles with ESM.


georg-schwarz:
@rhazn @joluj I'd appreciate on the configuration setup (esp. for vitest) before applying it to all libs/apps.

In particular, the `tsconfig.spec.ts` becomes unnecessary (?)

rhazn:
I don't know about the file (but if it doesn't break anything... :D). But adding .js to all imports has to be wrong, no? Especially since you have vitest imports in there that are without the .js addition. Also some import paths have /node_module/ in them, that also seems incorrect.

georg-schwarz:
> But adding .js to all imports has to be wrong, no?

I wish, but that is how ESM works: https://www.typescriptlang.org/docs/handbook/modules/reference.html

> Especially since you have vitest imports in there that are without the .js addition.

That is weird, true. Will investigate.

>  Also some import paths have /node_module/ in them, that also seems incorrect.

Thx, that must be incorrect.

PR #557, opened 2024-04-27
feat: enable transforming of value types

### LINKED ISSUE #543: [FEATURE] Enable transforming of value types
## User Story
1. As a {user}
2. I want {to change the value type of data}
3. So that {I can save it correctly}

## User Acceptance Criteria
- [ ] A `X as Y` operator exists that takes the input `X` and returns it parsed as built in value type `Y` (see https://jvalue.github.io/jayvee/docs/user/valuetypes/builtin-valuetypes)
  - [ ] Alternatively, multiple `asDecimal X` unary operators for every built in value type
- [ ] Transforms can define different value types for inputs and outputs (this should already work)
- [ ] `TableTransformer` blocks with transforms that change value types for the same columns work

## Examples
See the following model with open data from SWC: https://github.com/jvalue/SWC2-JValue-OCDE-Case1/blob/main/models/12781142/refractiveIndicesAndDielectricConstantsModel.jv

- The column `values` contains numeric data like this: `[123]`
- The goal is to save it as integer: `123`
- The `[RemoveClosingBraces](https://github.com/jvalue/SWC2-JValue-OCDE-Case1/blob/main/models/12781142/refractiveIndicesAndDielectricConstantsModel.jv#L53)` transform removes the braces (but returns the value as a `text` 😢 )
- The `[ClosingBracesRemover](https://github.com/jvalue/SWC2-JValue-OCDE-Case1/blob/main/models/12781142/refractiveIndicesAndDielectricConstantsModel.jv#L90)` block applies it to the pipeline
- The `[RemoveClosingBraces](https://github.com/jvalue/SWC2-JValue-OCDE-Case1/blob/main/models/12781142/refractiveIndicesAndDielectricConstantsModel.jv#L53)` transform should return the value as an integer using the following expression: `parse (valueWithBraces replace /\]/ with "") as integer`

## Definitions of Done
- [ ] All tests are passing

TungstnBallon:
This PR allows to parse `text` into the builtin primitives `decimal`, `integer` and `boolean`

example file:
```
pipeline CarsPipeline {

	CarsExtractor -> CarsTextFileInterpreter;

	CarsTextFileInterpreter
		-> CarsCSVInterpreter
		-> NameHeaderWriter
	   	-> CarsTableInterpreter
		-> DispTransform
		-> HpTransform
		-> CarsLoader;


		url: "https://gist.githubusercontent.com/noamross/e5d3e859aa0c794be10b/raw/b999fb4425b54c63cab088c0ce2c0d6ce961a563/cars.csv";
	}

	block CarsTextFileInterpreter oftype TextFileInterpreter { }

	block CarsCSVInterpreter oftype CSVInterpreter {
		enclosing: '"';
	}

	block NameHeaderWriter oftype CellWriter {
		at: cell A1;

		write: ["name"];
	}

	block CarsTableInterpreter oftype TableInterpreter {
		header: true;
		columns: [
			"name" oftype text,
			"mpg" oftype decimal,
			"cyl" oftype integer,
			"disp" oftype text,
			"hp" oftype text,
			"drat" oftype decimal,
			"wt" oftype decimal,
			"qsec" oftype decimal,
			"vs" oftype integer,
			"am" oftype integer,
			"gear" oftype integer,
			"carb" oftype integer
		];
	}

	transform parseDisp {
		from t oftype text;
		to d oftype decimal;
		d: asDecimal t;
	}

	block DispTransform oftype TableTransformer {
		inputColumns: ["disp"];
		outputColumn: "disp";
		use: parseDisp;
	}

	transform parseHp {
		from t oftype text;
		to i oftype integer;
		i: asInteger t;
	}

	block HpTransform oftype TableTransformer {
		inputColumns: ["hp"];
		outputColumn: "hp";
		use: parseHp;
	}

	block CarsLoader oftype SQLiteLoader {
		table: "Cars";
		file: "./cars.sqlite";
	}
}
```

closes #543

### DIFF
--- a/libs/language-server/src/lib/ast/expressions/evaluators/parse-operators-evaluators.spec.ts
+++ b/libs/language-server/src/lib/ast/expressions/evaluators/parse-operators-evaluators.spec.ts
@@ -1,0 +1,132 @@
+// SPDX-FileCopyrightText: 2024 Friedrich-Alexander-Universitat Erlangen-Nurnberg
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+import { executeDefaultTextToTextExpression } from '../test-utils';
+
+describe('The asText operator', () => {
+  it('should parse text successfully', async () => {
+    const result = await executeDefaultTextToTextExpression(
+      'asText inputValue',
+      'someText',
+    );
+
+    expect(result).toEqual('someText');
+  });
+});
+
+describe('The asDecimal operator', () => {
+  it('should parse positive decimals successfully', async () => {
+    const result = await executeDefaultTextToTextExpression(
+      'asDecimal inputValue',
+      '1.6',
+    );
+
+    expect(result).toEqual(1.6);
+  });
+
+  it('should parse decimals with commas successfully', async () => {
+    const result = await executeDefaultTextToTextExpression(
+      'asDecimal inputValue',
+      '1,6',
+    );
+
+    expect(result).toEqual(1.6);
+  });
+
+  it('should parse negative decimals with commas successfully', async () => {
+    const result = await executeDefaultTextToTextExpression(
+      'asDecimal inputValue',
+      '-1,6',
+    );
+
+    expect(result).toEqual(-1.6);
+  });
+});
+
+describe('The asInteger operator', () => {
+  it('should parse positive integers successfully', async () => {
+    const result = await executeDefaultTextToTextExpression(
+      'asInteger inputValue',
+      '32',
+    );
+
+    expect(result).toEqual(32);
+  });
+
+  it('should parse negative integers successfully', async () => {
+    const result = await executeDefaultTextToTextExpression(
+      'asInteger inputValue',
+      '-1',
+    );
+
+    expect(result).toEqual(-1);
+  });
+
+  it('should fail with decimal values', async () => {
+    const result = await executeDefaultTextToTextExpression(
+      'asInteger inputValue',
+      '32.5',
+    );
+
+    expect(result).toEqual(undefined);
+  });
+});
+
+describe('The asBoolean operator', () => {
+  it('should parse true and True successfully', async () => {
+    let result = await executeDefaultTextToTextExpression(
+      'asBoolean inputValue',
+      'true',
+    );
+
+    expect(result).toEqual(true);
+
+    result = await executeDefaultTextToTextExpression(
+      'asBoolean inputValue',
+      'True',
+    );
+
+    expect(result).toEqual(true);
+  });
+
+  it('should parse false and False successfully', async () => {
+    let result = await executeDefaultTextToTextExpression(
+      'asBoolean inputValue',
+      'false',
+    );
+
+    expect(result).toEqual(false);
+
+    result = await executeDefaultTextToTextExpression(
+      'asBoolean inputValue',
+      'False',
+    );
+
+    expect(result).toEqual(false);
+  });
+
+  it('should fail with 0 and 1', async () => {
+    let result = await executeDefaultTextToTextExpression(
+      'asBoolean inputValue',
+      '0',
+    );
+
+    expect(result).toEqual(undefined);
+
+    result = await executeDefaultTextToTextExpression(
+      'asBoolean inputValue',
+      '1',
+    );
+
+    expect(result).toEqual(undefined);
+  });
+
+  it('should fail on a random string', async () => {
+    const result = await executeDefaultTextToTextExpression(
+      'asBoolean inputValue',
+      'notABoolean',
+    );
+
+    expect(result).toEqual(undefined);
+  });
+});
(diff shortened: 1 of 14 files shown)


libs/language-server/src/lib/ast/expressions/evaluators/parse-operators-evaluators.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+    const result = await executeDefaultTextToTextExpression(
+      'asInteger inputValue',
+      '32',
+    );
+
+    expect(result).toEqual(32);
+  });
+});
+
+describe('The asInteger operator', () => {
```
georg-schwarz, line 28:
```suggestion
describe('The asBoolean operator', () => {
```

georg-schwarz (commented):
Cool, nice work!

Have you tried parsing something unparseable? What is the current behavior and does the user get an understandable log?

TungstnBallon:
> Cool, nice work!
 
Thanks :)

> Have you tried parsing something unparseable? What is the current behavior and does the user get an understandable log?

Rows where the parsing fails get excluded from the resulting table. The error is only visible with the `-d` flag.
e.g.
```
> nx run interpreter:run -d example/parse.jv

[CarsPipeline] Overview:
	Blocks (9 blocks with 2 pipes):
	-> OCarsExtractor (LocalFileExtractor)
		-> CarsTextFileInterpreter (TextFileInterpreter)
			-> CarsCSVInterpreter (CSVInterpreter)
				-> NameHeaderWriter (CellWriter)
					-> CarsTableInterpreter (TableInterpreter)
						-> DispTransform (TableTransformer)
							-> HpTransform (TableTransformer)
								-> CarsLoader (SQLiteLoader)

	[OCarsExtractor] Successfully extraced file /home/jonas/Downloads/parse.csv
	[OCarsExtractor] Execution duration: 1 ms.
	[CarsTextFileInterpreter] Decoding file content using encoding "utf-8"
	[CarsTextFileInterpreter] Splitting lines using line break /\r?\n/
	[CarsTextFileInterpreter] Lines were split successfully, the resulting text file has 33 lines
	[CarsTextFileInterpreter] Execution duration: 1 ms.
	[CarsCSVInterpreter] Parsing raw data as CSV using delimiter ","
	[CarsCSVInterpreter] Parsing raw data as CSV-sheet successful
	[CarsCSVInterpreter] Execution duration: 10 ms.
	[NameHeaderWriter] Writing "name" at cell A1
	[NameHeaderWriter] Execution duration: 1 ms.
	[CarsTableInterpreter] Matching header with provided column names
	[CarsTableInterpreter] Validating 32 row(s) according to the column types
	[CarsTableInterpreter] Validation completed, the resulting table has 32 row(s) and 12 column(s)
	[CarsTableInterpreter] Execution duration: 1 ms.
	[DispTransform] Column "disp" will be overwritten
	[DispTransform] Column "disp" will change its type from text to decimal
		[parseDisp] Invalid value in row 1: "NaN" does not match the type decimal
	[DispTransform] Execution duration: 1 ms.
	[HpTransform] Column "hp" will be overwritten
	[HpTransform] Column "hp" will change its type from text to integer
	[HpTransform] Execution duration: 1 ms.
	[CarsLoader] Opening database file ./cars.sqlite
	[CarsLoader] Dropping previous table "Cars" if it exists
	[CarsLoader] Creating table "Cars"
	[CarsLoader] Inserting 31 row(s) into table "Cars"
	[CarsLoader] The data was successfully loaded into the database
	[CarsLoader] Execution duration: 13 ms.
[CarsPipeline] Execution duration: 30 ms.
```

IMO the interpreter should't crash in this case, but a more visible error is necessary. I don't really know how to do this though, so some pointers would be welcome.

libs/language-server/src/lib/ast/expressions/evaluators/parse-operators-evaluators.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+      'asInteger inputValue',
+      '32',
+    );
+
+    expect(result).toEqual(32);
+  });
+});
+
+describe('The asBoolean operator', () => {
+  it('should parse integers successfully', async () => {
```
rhazn, line 29: A few more values would be great, e.g. 32.5 for integer, -1 for integer, '0' for boolean etc.

rhazn (changes requested):
Great work :D. I agree on not letting the interpreter crash, maybe we can change the error message to something more explicit like:  Invalid value in row 1: "NaN" does not match the type decimal -> "can not be cast to type decimal".

Otherwise, only blocking thing is the shared parsing logic.

TungstnBallon:
> Invalid value in row 1: "NaN" does not match the type decimal -> "can not be cast to type decimal".

The error message is now
```
[parsefailer] Could not parse "Mazda RX4" into a Decimal
[parsefailer] Dropping row 1: Could not evaluate transform expression
```

libs/language-server/src/lib/ast/expressions/evaluators/parse-operators-evaluators.ts
```diff
@@ -0,0 +1,70 @@
...
+  boolean
+> {
+  constructor(private readonly valueTypeProvider: ValueTypeProvider) {
+    super('asBoolean', STRING_TYPEGUARD);
+  }
+  override doEvaluate(operandValue: string): boolean {
+    const bool =
+      this.valueTypeProvider.Primitives.Boolean.fromString(operandValue);
+    if (bool === undefined) {
+      throw new Error(`Could not parse "${operandValue}" into an Boolean`);
```
rhazn, line 66:
```suggestion
      throw new Error(`Could not parse "${operandValue}" into a Boolean`);
```

rhazn (approved):
Top

PR #558, opened 2024-05-03
Jayvee goes ESM (Langium Upgrade)

georg-schwarz:


apps/language-server-web-worker/src/main.ts
```diff
@@ -3,12 +3,13 @@
...
 import { createJayveeServices } from '@jvalue/jayvee-language-server';
-import { EmptyFileSystem, startLanguageServer } from 'langium';
+import { EmptyFileSystem } from 'langium';
+import { startLanguageServer } from 'langium/lsp';
 import {
   BrowserMessageReader,
   BrowserMessageWriter,
   createConnection,
-} from 'vscode-languageserver/browser.js';
+} from 'vscode-languageserver/lib/browser/main.js';
```
georg-schwarz, line 12: Maybe add a comment why this is so weirdly imported.

rhazn (approved):

libs/execution/src/lib/blocks/composite-block-executor.ts
```diff
@@ -2,7 +2,7 @@
 //
 // SPDX-License-Identifier: AGPL-3.0-only
 
-import { strict as assert } from 'assert/strict';
+import { strict as assert } from 'node:assert/strict';
```
georg-schwarz, line 5: Import is off.

package.json
```diff
@@ -25,13 +25,14 @@
     "chalk": "^4.1.2",
     "clsx": "^1.2.1",
     "commander": "^8.0.0",
+    "eslint-plugin-vitest": "^0.5.4",
```
georg-schwarz, line 28: Should be a dev dependency

apps/docs-generator/src/main.ts
```diff
@@ -16,35 +17,33 @@ import { NodeFileSystem } from 'langium/node';
...
 import { UserDocGenerator } from './user-doc-generator';
 
+/** ESM does not know __filename and __dirname, so defined here */
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
 async function main(): Promise<void> {
-  const rootPath = join(__dirname, '..', '..', '..', '..');
+  const docsAppPath = join(__dirname, '..', '..', '..', 'apps', 'docs');
+  const jayveeExamplesPath = join(__dirname, '..', '..', '..', 'example');
```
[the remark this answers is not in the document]
georg-schwarz: We define it a few lines above ;-)
```
/** ESM does not know __filename and __dirname, so defined here */
const __filename = fileURLToPath(import.meta.url);
const __dirname = path.dirname(__filename);
```

joluj (approved):

Issue #563, opened 2024-05-14
[FEATURE] Multi-file Jayvee


georg-schwarz:
## User Story
1. As a {Jayvee programmer}
2. I want {to organize my program into multiple files}
3. So that {I can keep an overview in larger projects}

## User Acceptance Criteria
- [ ] [RFC 0015](https://github.com/jvalue/jayvee/blob/main/rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md) is implemented
  - [ ] `publish` and `use` functionality   
  - [ ] `package` functionality 

## Examples
See [RFC 0015](https://github.com/jvalue/jayvee/blob/main/rfc/0015-multi-file-jayvee/0015-multi-file-jayvee.md).

## Definitions of Done
- [ ] A PR has been opened and accepted
- [ ] All user acceptance criteria are met
- [ ] All tests are passing

PR #564, opened 2024-05-14
Simple file import

georg-schwarz:
Partially addresses #563: Introduces `use` functionality with wildcards to load elements from other files.

- wildcard use `use * from "relative/path/file.jv"`
- validation to ensure used files exist
- rename the `use` property of block type `TableTransformer` to `uses`, otherwise the new keyword `use` would be ambiguous
- for this PR, export all reusable root-level elements in a file besides pipelines
- when using the interpreter cli, always load the whole directory of the jv file into the workspace (so they can be used)
- For testing, introduce a helper function, `parseTestFileInWorkingDir` that allows loading the test files with their working directory instead of only reading the single file. This is a competing test setup with the used `parseHelper` function. We should discuss whether we need two separate test setups or if this one should be applied everywhere.

## Future Work

I tried to keep the PR short, so here an outlook of what's next

I see the follow-up tasks before tackling further import features:
- Auto-completion for file paths
- Rework of the test setup (as mentioned above)
- Check whether the import influences auto-completion functionality in general

I see the next import features to be:
- Explicitly `publish` elements instead of publishing all
- Explicitly `use` elements instead of using a wildcard


georg-schwarz:
@rhazn @joluj I'd like to tackle this PR in a test-driven manner. I added all test scenarios that came to my mind in [9637146](https://github.com/jvalue/jayvee/pull/564/commits/96371465aa06a0d685e09898b5704cb40334051b). Feedback on them is already very welcome ;-)

libs/language-server/src/lib/lsp/jayvee-scope-provider.ts
```diff
@@ -0,0 +1,99 @@
...
+   * Recursively add all imported URIs to @param importedUris
+   */
+  private gatherImports(
+    jayveeModel: JayveeModel,
+    importedUris: Set<string>,
+  ): void {
+    for (const importDefinition of jayveeModel.imports) {
+      const uri = this.importResover.resolveImportUri(importDefinition);
+      if (uri === undefined) {
+        break;
```
georg-schwarz, line 79: Here, all `break`s should be `continue`s I believe (?)

libs/interpreter-lib/src/parsing-util.spec.ts
```diff
@@ -8,65 +8,47 @@ import { CachedLogger, DiagnosticSeverity } from '@jvalue/jayvee-execution';
...
 
 import { extractDocumentFromFile, validateDocument } from './parsing-util';
 
 describe('Validation of parsing-util', () => {
-  let parse: (
-    input: string,
-    options?: ParseHelperOptions,
-  ) => Promise<LangiumDocument<AstNode>>;
-
+  const WORKING_DIR = path.resolve(__dirname, '../test/assets/parsing-util/');
```
joluj, line 22: Since we use esm now, shouldn't we restrain the use of `__dirname`?

libs/language-server/src/grammar/main.langium
```diff
@@ -25,3 +26,5 @@ entry JayveeModel:
   )*;
 
 
+ImportDefinition:
+  'use' '*' 'from' path=STRING ';';
```
joluj, line 30: This can later be refined to use named imports right?

libs/language-server/src/lib/validation/checks/import-definition.ts
```diff
@@ -0,0 +1,30 @@
+// SPDX-FileCopyrightText: 2023 Friedrich-Alexander-Universitat Erlangen-Nurnberg
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+
+import { type ImportDefinition } from '../../ast/generated/ast';
+import { type JayveeValidationProps } from '../validation-registry';
+
+/**
+ * See https://jvalue.github.io/jayvee/docs/dev/guides/working-with-the-ast/ for why the following ESLint rule is disabled for this file.
+ */
```
joluj, line 10: Leftover?

joluj (commented):
Looks good.

A test for transitive imports would be nice (A imports B, B imports C; Assert that A can use stuff from B).

libs/language-server/src/grammar/main.langium
georg-schwarz: Yes! That's the plan :smile:

libs/interpreter-lib/src/parsing-util.spec.ts
georg-schwarz: True! It seems to work, though, for the test. And it's used at some more places. I suggest to open a follow-up issue.

georg-schwarz: #568

joluj (approved):

libs/language-server/src/lib/lsp/import-reference.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const blocks = model.pipelines[0]!.blocks;
+    expect(blocks.length).toEqual(1); // pipeline contains one block
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const reference = blocks[0]!.type; // of an imported block type
+
+    expect(reference.ref).toBeDefined();
+  });
+
+  it('should resolve reference to transitively imported element', async () => {
```
rhazn, line 60: One of two things are happening with these tests:

1. I am too blind to understand what is tested here, to me the code for invalid and valid test cases looks exactly the same and the `expect` calls are also the same... so I fail to understand what is being tested here.
2. The tests are actually all the same except for the input file.

In case 1, I think the test code could be refactored so it is more obvious what is being tested exactly. In case 2, the tests should be fixed.

libs/language-server/src/lib/lsp/jayvee-scope-computation.ts
```diff
@@ -0,0 +1,52 @@
...
+    // export the exportable top-level elements
+    if (!this.isExportable(node)) {
+      return;
+    }
+
+    super.exportNode(node, exports, document);
+  }
+
+  isExportable(node: AstNode) {
+    // pipelines are not exported
```
rhazn, line 41: If the logic is pipelines are not exported, why not disallow list them instead of allowlisting everything else? Would be better for future additions to the language.

libs/language-server/src/lib/validation/checks/import-definition.ts
```diff
@@ -0,0 +1,26 @@
...
+  checkPathExists(importDefinition, props);
+}
+
+function checkPathExists(
+  importDefinition: ImportDefinition,
+  props: JayveeValidationProps,
+): void {
+  const resolvedImport = props.importResolver.resolveImport(importDefinition);
+  if (resolvedImport === undefined) {
+    props.validationContext.accept('error', 'Import cannot be resolved.', {
```
rhazn, line 21: Would be stronger as `Import from "{path}" could not be resolved. File does not exist or is not readable.` (if that is actually true / possible). Having the path in there will be very helpful for future debugging.

rhazn (approved):
Amazing. 👍

libs/language-server/src/lib/lsp/jayvee-scope-computation.ts
georg-schwarz: I agree, but then comments (hidden nodes) are exported as well.
Then we'd have to find out what to not export first I guess.
The more I think about it, the more reasonable it seems. Will give it a try!

libs/language-server/src/lib/lsp/import-reference.spec.ts
georg-schwarz: Yes, the tests are pretty similar. The input files should be specific to the test case, though.

For the valid cases, the reference should be followable (`.ref` is **not** `undefined`).
For the invalid cases, the reference should not be followable  (`.ref` **is** `undefined`).

I'll try to refactor using `test.each`, so we just modify the file path for every valid/invalid case. So there will be one test case for valid and one for invalid examples.

rhazn: Ah yeah, then it is a case 1 issue. In my defense, the only difference then is the `Un` in `toBeUndefined` :D. I think refactoring it into one method that just takes a file path and undefined/defined expectation as param would make this much more readable yeah.

georg-schwarz: I agree, hard to spot the difference :)

libs/language-server/src/lib/lsp/jayvee-scope-computation.ts
georg-schwarz: I gave it some more thought and came to the conclusion that we should leave it as it is.

Your suggestion makes total sense, but it will be solved by the next update anyway that makes exports explicit (`publish` keyword). So I'd not spend more effort than necessary as it will be reworked anyway.

libs/language-server/src/lib/validation/checks/import-definition.ts
georg-schwarz: Made the error message a little more vague by saying "Check if the file exists".
Tackled in [63c607b](https://github.com/jvalue/jayvee/pull/564/commits/63c607b9370100465427fe2b5c3cd47c90612211)

libs/language-server/src/lib/lsp/import-reference.spec.ts
georg-schwarz: Done in [ad744b7](https://github.com/jvalue/jayvee/pull/564/commits/ad744b78dcef3bb1bf5ac7dd4c8c77a361322343)

PR #571, opened 2024-05-23
Add `publish` keyword

georg-schwarz:
Partially addresses https://github.com/jvalue/jayvee/issues/563: Introduces `publish` keyword.
- Extended the grammar with the keyword
- Adjusted scoping to only make published elements available in other files
- Added validation to avoid cyclic imports
- Published all builtin elements
- Adapted tests

As discussed in https://github.com/eclipse-langium/langium/discussions/1508#discussioncomment-9524544, the main logic resides in the `JayveeScopeProvider` instead of the `JayveeScopeComputation`.

## Notes
- [x] Decide whether we should add caching (as suggested in the thread) right away or track it in a separate issue: already implemented, only 3 lines of code.
- [x] Validate code actions of #565 still work
- [x] Validate that cyclic dependencies are prevented
- [x] Add user documentation


libs/language-server/src/lib/lsp/jayvee-scope-provider.ts
```diff
@@ -40,22 +52,103 @@ export class JayveeScopeProvider extends DefaultScopeProvider {
...
+        }
+      }
+    }
+    return exportedElements;
+  }
+
+  /**
+   * Follow an export statement to its original definition.
+   */
+  protected followExportDefinitionChain(
```
georg-schwarz, line 111: If we should ever allow `publish * from '{path}'`, we'd have to use recursion here I think to follow the chain.. Right now, the exported element is directly referenced.

libs/language-server/src/lib/lsp/jayvee-scope-provider.ts
```diff
@@ -40,22 +52,103 @@ export class JayveeScopeProvider extends DefaultScopeProvider {
       return EMPTY_SCOPE;
     }
 
+    // TODO: add caching to avoid performance issues
```
georg-schwarz, line 55: Not sure whether to add in this PR or leave for future optimization.

libs/execution/src/lib/constraints/executors/range-constraint-executor.spec.ts
```diff
@@ -52,10 +53,15 @@ describe('Validation of RangeConstraintExecutor', () => {
       document.parseResult.value,
       'pipelines@0/blocks@2',
     ) as BlockDefinition;
-    const constraint = locator.getAstNode<TypedConstraintDefinition>(
-      document.parseResult.value,
-      'constraints@0',
-    ) as TypedConstraintDefinition;
+
+    const allElements = AstUtils.streamAllContents(document.parseResult.value);
```
rhazn, line 57: Can be merged like this but the code is repeated a lot. Much better would be to write a `expectConstraintsToExist` util method, put it into a test util file and just call it everywhere.

libs/language-server/src/lib/ast/expressions/test-utils.ts
```diff
@@ -45,10 +46,14 @@ export async function executeExpressionTestHelper(
...
     `);
 
-  const transform = locator.getAstNode<TransformDefinition>(
-    document.parseResult.value,
-    'transforms@0',
-  ) as TransformDefinition;
+  const allElements = AstUtils.streamAllContents(document.parseResult.value);
+  const allTransforms = [...allElements.filter(isTransformDefinition)];
+  expect(
+    allTransforms.length !== 0,
```
rhazn, line 52: Test and comment are at adds, test is true for 2, comment wants exactly 1.

rhazn (approved):
Had a high level look, overall looks good to me, great work 👍 . Just some small codestyle / optional comments.

libs/language-server/src/lib/ast/expressions/test-utils.ts
georg-schwarz: Nice catch! That line is actually wrong like this and does nothing. Fixed it in [3db1d0e](https://github.com/jvalue/jayvee/pull/571/commits/3db1d0ea9fced4a107c23d1fe0e0562b2ed523eb)

libs/execution/src/lib/constraints/executors/range-constraint-executor.spec.ts
georg-schwarz: On the one hand, I agree that the code repeats quite a bit here. On the other hand, it is test-case specific test setup. Generalizing here might introduce coupling between tests...

So I ended up creating a helper function that is flexible by passing a customizable filter function as parameter.

PR #573, opened 2024-05-27
Publish via alias

georg-schwarz:
Builds on #571. Introduces `export Element as Alias;` syntax.

### DIFF
--- a/libs/language-server/src/lib/validation/checks/export-definition.ts
+++ b/libs/language-server/src/lib/validation/checks/export-definition.ts
@@ -1,0 +1,134 @@
+/* eslint-disable @typescript-eslint/no-unnecessary-condition */
+// SPDX-FileCopyrightText: 2023 Friedrich-Alexander-Universitat Erlangen-Nurnberg
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+
+// eslint-disable-next-line unicorn/prefer-node-protocol
+import { strict as assert } from 'assert';
+
+import { AstUtils } from 'langium';
+
+import {
+  type ExportDefinition,
+  type ExportableElement,
+  type JayveeModel,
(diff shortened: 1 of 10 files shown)


libs/language-server/src/lib/validation/checks/export-definition.ts
```diff
@@ -0,0 +1,134 @@
...
+  const model = AstUtils.getContainerOfType(exportDefinition, isJayveeModel);
+  assert(model !== undefined);
+  const allExports = collectAllExportsWithinSameFile(model);
+
+  const elementsWithSameName = allExports.filter(
+    (e) => e.alias === exportDefinition.alias,
+  );
+  assert(
+    elementsWithSameName.length > 0,
+    'Could not the export definition itself in exports',
```
georg-schwarz, line 75:
```suggestion
    'Could not find the export definition itself in exports',
```

libs/language-server/src/lib/validation/checks/export-definition.ts
```diff
@@ -0,0 +1,134 @@
...
+      `This alias is ambiguous. There is another element published element with the same name "${exportDefinition.alias}".`,
+      {
+        node: exportDefinition,
+        property: 'alias',
+      },
+    );
+  }
+}
+
+function collectAllExportsWithinSameFile(model: JayveeModel): {
```
georg-schwarz, line 91: Should use a util function IMO

rhazn (approved):

libs/language-server/src/lib/validation/checks/export-definition.ts
georg-schwarz: Will be part of #574 as the logic is refactored there anyway.

PR #574, opened 2024-05-28
Import Specific Element

georg-schwarz:
Builds on #573. Diff: https://github.com/jvalue/jayvee/compare/publish-alias...import-specific-element?expand=1


libs/language-server/src/lib/lsp/jayvee-scope-provider.ts
```diff
@@ -68,114 +55,103 @@ export class JayveeScopeProvider extends DefaultScopeProvider {
...
-        exportedElements.push({
-          element: node,
-          alias: node.name,
-        });
-      }
+  protected getImportedElementsFromDocument(
+    importDefinition: ImportDefinition,
+    importedDocument: LangiumDocument,
+  ): AstNodeDescription[] {
+    const publishedElement =
```
rhazn, line 89: Nitpicky but the variable sounds like one element, the function like a list of elements.

rhazn (approved):

PR #580, opened 2024-05-31
Cache task "generate"

joluj:
This PR introduces the caching of the task `language-server:generate`.

Current state (before this PR): Langium code is generated on every build of the language-server (and due to library dependencies on every build of the interpreter and other apps). This slows down development.

With changes: `language-server:generate` detects if relevant files are changed and only runs when necessary.


rhazn (approved):

joluj:
I'm waiting for @georg-schwarz's feedback as well before merging.

georg-schwarz (changes requested):
Nice idea!

In an ideal world we'd define all elements in Jayvee itself, but we are not there yet. If you look into the stdlib generator you'll see that some parts of the stdlib are still generated from code, e.g., 

```
'builtin:///stdlib/io-types.jv': Object.values(IOType)
    .map((iotype) => `publish builtin iotype ${iotype};`)
    .join('\n\n'),
```

When changing the IOType enum, the caching should respect that as well. More difficult will be the built-in value types... Probably add the whole directory there.

If you add those things so that caching gets invalidated on their file changes, I'm fine with merging :)

joluj:
If you change `IOType` or `CSVExtractor.js` (built-in Javee Block), the `language-server` lib is rebuilt / its cache is invalidated. This is how it should be right? So everything works as expected

georg-schwarz (approved):

PR #583, opened 2024-06-05
docs: :memo: Add cell selection syntax docs

### LINKED ISSUE #582: [FEATURE] Cell Range Syntax Documentation
## User Story
1. As a {user}
2. I want {to be able to read how the cell selection syntax works}
3. So that {I can use blocks like the RowDeleter}

## User Acceptance Criteria
- [ ] Cell selection syntax is documented

## Definitions of Done
- [ ] All tests are passing

rhazn:
closes #582 

syntax-docs


apps/docs/docs/user/cell-range.md
```diff
@@ -0,0 +1,243 @@
+---
+sidebar_position: 9
+---
+
+# Cell Range
+
+Cell ranges can be used to select a subset of cells in a 2D data structure, such as a `Sheet`. The syntax for cell ranges follows conventions from spreadsheet-software. Rows are referenced by one-based indices and columns by capitalized characters, ordered alphabetically.
```
georg-schwarz, line 7: "such as sheet" implies that the syntax would also be usable for other 2D data structures, such as tables. Right now, we only allow it on sheets, right? Could make sense whether we want the same syntax for tables or only for sheets.

That is however unrelated to the changes. Just wanted to bring it up as it came to my mind :)

georg-schwarz (approved):

apps/docs/docs/user/cell-range.md
rhazn: True, even though tables always have column names so I assume we'd use those. But I could see other 2D data structures in the future (since sheets are explicitly strings right now, maybe we have one with individual cells having data types), so I wanted to keep the wording future proof 😅 .

Issue #587, opened 2024-06-11
[FEATURE] VSCode Extension: Run pipeline button


georg-schwarz:
## User Story
1. As a {VSCode user}
2. I want {just to press a button to run a Jayvee pipeline}
3. So that {I don't need to do it via the command line myself}

## User Acceptance Criteria
- [ ] Pipelines are runnable individually by the interpreter
- [ ] The VSCode extension checks whether there is a correct version of the Jayvee CLI installed
- [ ] There is a button for users to press close to the pipeline definition
  - [ ] Clicking the button leads to an automated execution of the pipeline 

## Notes
CodeLenses might be a LSP native feature to implement this.

## Definitions of Done
- [ ] A PR has been opened and accepted
- [ ] All user acceptance criteria are met
- [ ] All tests are passing

PR #592, opened 2024-06-19
Add CLI flag to select pipelines to execute

georg-schwarz:
Required for https://github.com/jvalue/jayvee/issues/587

### DIFF
--- a/apps/interpreter/src/index.ts
+++ b/apps/interpreter/src/index.ts
@@ -62,6 +62,11 @@ program
     'Only parses the model without running it. Exits with 0 if the model is valid, with 1 otherwise.',
     false,
   )
+  .option(
+    '-p, --pipeline <regular expression to match with pipeline name>',
+    'Only runs the matching pipelines (matching by name) and ignores other pipelines in the file.',
+    '.*',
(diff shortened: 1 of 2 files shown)


rhazn (approved):

rhazn:
Ah failed to comment on the review on Mobile, but: I think the approach is weird to make it a required parameter with a default value. I think it would be better as an optional parameter and if it's undefined, run all pipelines. But I don't mind that much, so your choice.

joluj (approved):

joluj:
> I think the approach is weird to make it a required parameter with a default value. 

Isn't that often like that? The default is running all pipelines, so I think that's fine.
It saves us a bit of complexity to know that the string cannot be `undefined`

rhazn:
Yeah, that's why I don't mind as much. But it will lead to some weird errors down the line when you write "can't find a pipeline for .*" and the user is confused because they never entered that. I know that is not how it would work right now, but by giving it a default value and then using that value later on in error messages, this kind of stuff can happen.

But as I said, I think that is a real edge case and I am happy with this as well.

PR #594, opened 2024-06-20
Refactor Interpreter (to an Interface + Implementation Class)

georg-schwarz:
- reduces amounts of parameters in the interpretation model by using object construction
- defines a nicer interpreter interface that is easier to use
- pulls out CLI runtime argument validation (should not be located in interpreter-lib)
- introduces a default value for the `--debug-target` parameter to align with other params: use a default value over undefined

### DIFF
--- a/apps/interpreter/src/run-action.ts
+++ b/apps/interpreter/src/run-action.ts
@@ -4,7 +4,6 @@
 
 import process from 'node:process';
 
-import { type Logger } from '@jvalue/jayvee-execution';
 import {
   DefaultJayveeInterpreter,
   ExitCode,
@@ -17,7 +16,7 @@ import {
   type JayveeServices,
 } from '@jvalue/jayvee-language-server';
 
-import { parseRunOptions } from './run-options';
+import { parsePipelineMatcherRegExp, parseRunOptions } from './run-options';
 
 export async function runAction(
   filePath: string,
@@ -65,19 +64,4 @@ async function runParseOnly(
   );
   const exitCode = model === undefined ? ExitCode.FAILURE : ExitCode.SUCCESS;
   process.exit(exitCode);
-}
-
-function parsePipelineMatcherRegExp(
-  matcher: string,
-  logger: Logger,
-): RegExp | undefined {
-  try {
-    return new RegExp(matcher);
-  } catch (e: unknown) {
-    logger.logErr(
-      `Invalid value "${matcher}" for pipeline selection option: -p --pipeline.\n` +
-        'Must be a valid regular expression.',
-    );
-    return undefined;
-  }
 }
--- a/apps/interpreter/src/run-action.ts
+++ b/apps/interpreter/src/run-action.ts
@@ -2,13 +2,23 @@
 //
 // SPDX-License-Identifier: AGPL-3.0-only
 
+// eslint-disable-next-line unicorn/prefer-node-protocol
+import { strict as assert } from 'assert';
 import process from 'node:process';
 
+import {
+  type DebugGranularity,
+  DebugGranularityValues,
+  type DebugTargets,
+  DefaultDebugTargetsValue,
+  type Logger,
+  isDebugGranularity,
+} from '@jvalue/jayvee-execution';
 import {
   DefaultJayveeInterpreter,
   ExitCode,
   type JayveeInterpreter,
-  type LoggerFactory,
+  LoggerFactory,
   extractAstNodeFromFile,
 } from '@jvalue/jayvee-interpreter-lib';
 import {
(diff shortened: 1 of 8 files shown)


joluj:
> This branch cannot be rebased due to conflicts

:(

apps/interpreter/src/run-options.ts
```diff
@@ -0,0 +1,181 @@
+// SPDX-FileCopyrightText: 2023 Friedrich-Alexander-Universitat Erlangen-Nurnberg
```
joluj, line 1: Since you've moved things to a class, does it make sense to continue doing that with the run options as well? They all belong together.

Also, it might make sense to move `parsePipelineMatcherRegExp` there.

apps/interpreter/src/run-action.ts
```diff
@@ -4,36 +4,80 @@
...
+        filePath,
+        services,
+        loggerFactory.createLogger(),
+      ),
+  );
+  const exitCode = model === undefined ? ExitCode.FAILURE : ExitCode.SUCCESS;
+  process.exit(exitCode);
+}
+
+function parsePipelineMatcherRegExp(
```
joluj, line 70: Doesn't this function make more sense in `run-options.ts`?

apps/interpreter/src/run-options.ts
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  const options = optionsRaw as {
+    pipeline: unknown;
+    env: unknown;
+    debug: unknown;
+    debugGranularity: unknown;
+    debugTarget: unknown;
+    parseOnly: unknown;
+  };
```
joluj, line 58:
```suggestion
  const options = optionsRaw as Record<keyof RunOptions, unknown>;
```

This way we don't require any changes here when we change `RunOptions`.

apps/interpreter/src/run-options.ts
```diff
@@ -0,0 +1,181 @@
...
+
+  isPipelineArgument(options.pipeline, logger);
+  isEnvArgument(options.env, logger);
+  isDebugArgument(options.debug, logger);
+  isDebugGranularityArgument(options.debugGranularity, logger);
+  isDebugTargetArgument(options.debugTarget, logger);
+  isParseOnlyArgument(options.parseOnly, logger);
+
+  // TypeScript does not infer type from type guards, probably fixed in TS 5.5
```
joluj, line 67: Well you don't check the type guards :D You need a condition

```
if (!isPipelineArgument(...)) {
  return null;
}

// Now options.pipeline should be the correct type
```

libs/interpreter-lib/src/interpreter.ts
```diff
@@ -2,19 +2,19 @@
 //
 // SPDX-License-Identifier: AGPL-3.0-only
 
-// eslint-disable-next-line unicorn/prefer-node-protocol
-import { strict as assert } from 'assert';
+// eslint-disable-next-line no-restricted-imports
```
joluj, line 5: What does that error mean?

joluj (changes requested):
The type guards currently don't do anything, this needs to be changed. The other stuff is optional.

apps/interpreter/src/run-options.ts
georg-schwarz: Wtf was I doing yesterday :rofl:

libs/interpreter-lib/src/interpreter.ts
georg-schwarz: Thx, nice catch. Disabled the wrong linter rule here :)

apps/interpreter/src/run-options.ts
georg-schwarz: Do you mean turning the `RunOptions` from an interface to a class?

joluj: Rather move all of these helper functions to a class

georg-schwarz: In this specific case, I don't see the big benefit compare to a collection of methods. We can refactor it when we really need it.

For the interpreter, I saw a larger benefit as we had methods with 5+ parameters. It should also pave the way to configure the workspace directory more easily.

If you don't insist on that change, I'd skip it 😅

joluj: sure lets skip it :+1:

apps/interpreter/src/run-options.ts
```diff
@@ -48,29 +48,26 @@ export function parseRunOptions(
...
+    !isEnvArgument(options.env, logger) ||
+    !isDebugArgument(options.debug, logger) ||
+    !isDebugGranularityArgument(options.debugGranularity, logger) ||
+    !isDebugTargetArgument(options.debugTarget, logger) ||
+    !isParseOnlyArgument(options.parseOnly, logger)
+  ) {
+    return undefined;
+  }
 
   // TypeScript does not infer type from type guards, probably fixed in TS 5.5
```
joluj, line 64: This comment should be removed

joluj (approved):
That one comment does not make sense anymore. Otherwise looks good

PR #595, opened 2024-06-21
Incorporate missing change proposals to RFC 00014

georg-schwarz:
I think we didn't integrate some change proposals we agreed on. This PR introduces those changes.

### DIFF
--- a/rfc/0014-attribute-based-valuetypes/README.md
+++ b/rfc/0014-attribute-based-valuetypes/README.md
@@ -9,7 +9,7 @@ SPDX-License-Identifier: AGPL-3.0-only
 |             |                               |
 | ----------- | ----------------------------- | --------------------------------------------------------------- |
 | Feature Tag | `attribute-based value types` | <!-- TODO: choose a unique and declarative feature name -->     |
-| Status      | `DISCUSSION`                       | <!-- Possible values: DRAFT, DISCUSSION, ACCEPTED, REJECTED --> |
+| Status      | `DISCUSSION`                  | <!-- Possible values: DRAFT, DISCUSSION, ACCEPTED, REJECTED --> |
 | Responsible | `dirkriehle`                  | <!-- TODO: assign yourself as main driver of this RFC -->       |
 
 <!--
@@ -42,10 +42,10 @@ This syntax is smart in that you don't have to list and name an attribute but ra
 The current syntax, in which an underlying value type is referenced through 'oftype', it does not use instantiation/inheritance but rather composition.
 The purpose of using composition is that it is
 
-1. more similar to traditional approaches and
+1. more similar to traditional programming and
 2. is needed anyway for multi-attribute value types.
 
-Thus, this RFC removes the `oftype` keyword in a value type definitions and proposes an attribute-based syntax that is extendable to a future multi-attribute syntax.
+Thus, this RFC removes reference to a parent value type in a value type definition and proposes an explicit attribute-based syntax instead that is extendable to a future multi-attribute syntax.
 
 ## Explanation
 
@@ -55,14 +55,14 @@ I propose to make that attribute explicit. The new syntax would be:
 valuetype CorrelationCoefficient {
   property correlation oftype decimal;
 
-  constraint minusOneToPlusOneRange: MinusOneToPlusOneRange(value = correlation);
+  constraint minusOneToPlusOneRange: MinusOneToPlusOneRange on correlation;
 }
 
 constraint MinusOneToPlusOneRange on decimal:
   value >= -1 and value <= 1;
 ```
 
-The syntax is kept flat rather than using arrays.
+The syntax is kept flat rather than using arrays in alignment with the [Jayvee Design Principles](https://jvalue.github.io/jayvee/docs/0.4.0/dev/design-principles#jayvee-manifesto).
 
 Despite referencing reusable constraints, the syntax also allows in-place constraint definition:
 
@@ -71,7 +71,7 @@ valuetype CorrelationCoefficient {
   property correlation oftype decimal;
 
   constraint minusOneToPlusOneRange:
-    correlation >= 1 and correlation <=1;
+    correlation >= -1 and correlation <=1;
 }
 ```
 
--- a/rfc/0014-attribute-based-valuetypes/README.md
+++ b/rfc/0014-attribute-based-valuetypes/README.md
@@ -71,7 +71,7 @@ valuetype CorrelationCoefficient {
   property correlation oftype decimal;
 
   constraint minusOneToPlusOneRange:
-    correlation >= -1 and correlation <=1;
+    correlation >= -1 and correlation <= 1;
 }
 ```
 
--- a/rfc/0014-attribute-based-valuetypes/README.md
+++ b/rfc/0014-attribute-based-valuetypes/README.md
@@ -6,12 +6,13 @@ SPDX-License-Identifier: AGPL-3.0-only
 
 # RFC 0014: Attributed-based syntax for defining value types
 
-| | |
-|---|---|
-| Feature Tag | `attribute-based value types` | <!-- TODO: choose a unique and declarative feature name -->
-| Status | `DRAFT` | <!-- Possible values: DRAFT, DISCUSSION, ACCEPTED, REJECTED -->
-| Responsible | `dirkriehle` | <!-- TODO: assign yourself as main driver of this RFC -->
-<!-- 
+|             |                               |
+| ----------- | ----------------------------- | --------------------------------------------------------------- |
+| Feature Tag | `attribute-based value types` | <!-- TODO: choose a unique and declarative feature name -->     |
+| Status      | `DISCUSSION`                       | <!-- Possible values: DRAFT, DISCUSSION, ACCEPTED, REJECTED --> |
+| Responsible | `dirkriehle`                  | <!-- TODO: assign yourself as main driver of this RFC -->       |
+
+<!--
   Status Overview:
   - DRAFT: The RFC is not ready for a review and currently under change. Feel free to already ask for feedback on the structure and contents at this stage.
   - DISCUSSION: The RFC is open for discussion. Usually, we open a PR to trigger discussions.
@@ -21,16 +22,10 @@ SPDX-License-Identifier: AGPL-3.0-only
 
 ## Summary
 
-I'd like to allow the definition of user-defined single-attribute value types. In addition to the current syntax, in which an underlying value type is referenced through 'oftype', it does not use instantiation/inheritance but rather composition. 
+Value types are a core concept in Jayvee for data validation.
+This RFC reworks the syntax to an attribute-based syntax.
 
 ## Motivation
-
-The purpose of using composition is that it is
-
-1. more similar to traditional approaches and
-2. is needed anyway for multi-attribute value types.
-
-## Explanation
 
 To define a value type like CorrelationCoefficient and its range of -1 to +1, we have to write:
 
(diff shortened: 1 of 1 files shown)


rfc/0014-attribute-based-valuetypes/README.md
```diff
@@ -42,47 +37,61 @@ valuetype CorrelationCoefficient oftype decimal {
...
 valuetype CorrelationCoefficient {
-  attributes: [
-    value oftype decimal;
-  ];
-  constraints: [
-    MinusOneToPlusOneRange; // Don't know how to attach this to value attribute
-  ];
+  property correlation oftype decimal;
+
+  constraint minusOneToPlusOneRange: MinusOneToPlusOneRange(value = correlation);
```
georg-schwarz, line 58: Not entirely sure if we really agreed on the syntax to call a constraint on a property

rfc/0014-attribute-based-valuetypes/README.md
```diff
@@ -42,47 +37,61 @@ valuetype CorrelationCoefficient oftype decimal {
...
-This syntax is smart in that you don't have to list and name an attribute but rather rely on an implicit 'value' attribute. Still, I propose to make that attribute explicit. New syntax would be:
+This syntax is smart in that you don't have to list and name an attribute but rather rely on an implicit 'value' attribute.
+
+The current syntax, in which an underlying value type is referenced through 'oftype', it does not use instantiation/inheritance but rather composition.
+The purpose of using composition is that it is
+
+1. more similar to traditional approaches and
+2. is needed anyway for multi-attribute value types.
+
+Thus, this RFC removes the `oftype` keyword in a value type definitions and proposes an attribute-based syntax that is extendable to a future multi-attribute syntax.
```
rhazn, line 48:
```suggestion
Thus, this RFC removes the reference to a parent value type in a value type definition and proposes an explicit attribute-based syntax instead that is extendable to a future multi-attribute syntax.
```

rfc/0014-attribute-based-valuetypes/README.md
rhazn: I don't think we did (and I don't think we have a `constraint` keyword somewhere that is not a constraint definition, right? I'd keep the suggestions in this PR more minimal with 

```
constraints: [
  MinusOneToPlusOneRange on correlation
]
```

I think that nicely aligns with our current syntax style (so doesn't introduce new concepts like constructor-like brackets and assignments). We will of course have to evolve on that (e.g. constraints over multiple attributes, AND/ORing constraints etc) but I'd do that in a later RFC.

rfc/0014-attribute-based-valuetypes/README.md
```diff
@@ -42,47 +37,61 @@ valuetype CorrelationCoefficient oftype decimal {
...
+
+  constraint minusOneToPlusOneRange: MinusOneToPlusOneRange(value = correlation);
 }
+
+constraint MinusOneToPlusOneRange on decimal:
+  value >= -1 and value <= 1;
 ```
 
-While more verbose, it prepares the way for 
+The syntax is kept flat rather than using arrays.
```
rhazn, line 65:
```suggestion
The syntax is kept flat rather than using arrays in alignment with the [Jayvee Design Principles](https://jvalue.github.io/jayvee/docs/0.4.0/dev/design-principles#jayvee-manifesto).
```

rfc/0014-attribute-based-valuetypes/README.md
```diff
@@ -42,47 +37,61 @@ valuetype CorrelationCoefficient oftype decimal {
...
+
+Despite referencing reusable constraints, the syntax also allows in-place constraint definition:
```
rhazn, line 67: I would keep this out of the RFC to make it smaller. We should probably have in-place constraint definitions but imho can do so in a follow-up.

rhazn (approved):

rfc/0014-attribute-based-valuetypes/README.md
```diff
@@ -42,47 +37,61 @@ valuetype CorrelationCoefficient oftype decimal {
 }
 ```
 
-This syntax is smart in that you don't have to list and name an attribute but rather rely on an implicit 'value' attribute. Still, I propose to make that attribute explicit. New syntax would be:
+This syntax is smart in that you don't have to list and name an attribute but rather rely on an implicit 'value' attribute.
+
+The current syntax, in which an underlying value type is referenced through 'oftype', it does not use instantiation/inheritance but rather composition.
+The purpose of using composition is that it is
+
+1. more similar to traditional approaches and
```
joluj, line 45: What is a traditional approach? Since you mentioned inheritance before, does it correlate to that?

rfc/0014-attribute-based-valuetypes/README.md
joluj: This sentence does not make sense anymore if we apply @rhazn's suggestion [above](https://github.com/jvalue/jayvee/pull/595/files#r1650751505).

rfc/0014-attribute-based-valuetypes/README.md
joluj: I wouldn't mind adding it here. It's not huge anyway.

rfc/0014-attribute-based-valuetypes/README.md
```diff
@@ -42,47 +37,61 @@ valuetype CorrelationCoefficient oftype decimal {
...
-valuetype Money {
-  attributes: [
-    amount oftype decimal;
-    currency oftype Currency;
-  ];
+valuetype CorrelationCoefficient {
+  property correlation oftype decimal;
+
+  constraint minusOneToPlusOneRange:
+    correlation >= 1 and correlation <=1;
```
joluj, line 74:
```suggestion
    correlation >= -1 and correlation <= 1;
```

rfc/0014-attribute-based-valuetypes/README.md
```diff
@@ -42,47 +37,61 @@ valuetype CorrelationCoefficient oftype decimal {
...
-2. Creates extra work/may be too difficult, and
-3. May be disruptive to how the language currently works. 
+1. No backward compatible syntax (breaking change).
+2. Removes the inheritance mechanism.
 
 ## Alternatives
 
-Conceivably, we could use multiple inheritance for multi-attribute value types... just joking. 
+1. Keep the current syntax.
+2. Use arrays instead of the flat syntax.
```
joluj, line 86: Remember to remove this if we apply @rhazn's [suggestion](https://github.com/jvalue/jayvee/pull/595/files#r1650751505)

joluj (commented):
I'm not really happy with the new syntax regarding the constraints. I like @rhazn's suggestion more, but I'm not really happy with that either.
But I think we will get better ideas when we use it a bit and when we think about AND, OR, XOR, ...

Since we need to go over the constraints later anyway, I'm fine with not finding an optimal solution now.

rfc/0014-attribute-based-valuetypes/README.md
rhazn: Just for a view on the syntax, if we go with my suggestion, that would be:

```
constraints: [ value >= 1 and value <=1 on correlation ]
```

equivalent to

```
constraints: [ MinusOneToPlusOneRange on correlation ]
```

My issue with that was that reading the in-place constraint definition with the following "on" is kind of confusing... But honestly, seeing how it looks here I am kind of happy with it. So I'll revert what I said and say we should have in place constraints as well, yeah.

joluj: Why shouldn't we use
```
constraints: [ correlation >= 1 and correlation <=1 ]
```
?

And combined:
```
constraints: [
    correlation >= 1 and correlation <=1,
    MinusOneToPlusOneRange on correlation
]
```
I don't see why we need to use `on` for in-place constraints as well

joluj: I use `correlation` instead of `value`.
This allows for example
```
constraints: [
    correlation >= 1 and otherUnit <=1
]
```

rhazn: Ah, fair enough. I dislike that because it introduces different syntax for inline constraints vs. references and I like the ability to always just create a constraint reference by c&p the inline code. But that is something that reasonable people can disagree on :D.

rfc/0014-attribute-based-valuetypes/README.md
georg-schwarz: How about the following?

```
constraint MinusOneToPlusOneRange on correlation;
```

I'd really like to get rid of the array syntax and more closely align with the "flat over nested" design principle.

rfc/0014-attribute-based-valuetypes/README.md
georg-schwarz: I see your point @rhazn, and I see value in keeping the syntax consistent independently if defined in-line or as stand-alone element.

What do you think about changing the **constraint definition** syntax in a similar way to keep it consistent? Instead of the `value` keyword, constraint definitions would refer to the value type's properties they are based on:

```
constraint CorrelationCoefficient on decimal:
  value >= 1 and value <= 1; // "value" is the property name of builtin value types (so nothing changes for them)
  
constraint PostiveCorrelationConstraint on CorrelationCoefficient:
  correlation > 0; // "correlation" is a property of the value type CorrelationCoefficient
  
// Future, for multi-dimensional value types
constraint Coordinate2DInLeftUpperQuadrantConstraint on Coordinate2D:
  x > 0 and y > 0;
```

An obvious downside is the circular dependency of value type and constraint that could be incentivized (or should be prevented by hinting to in-line constraints in such cases).

rfc/0014-attribute-based-valuetypes/README.md
rhazn: You mean similar to the `property` syntax of just listing all the properties in that way? I am good with it, happy with your suggestion there.

georg-schwarz: Yes exactly. I feel like, currently, we have two different approaches to defining lists. And I'd like to introduce more consistency by also using the same flat syntax as for `property`

rfc/0014-attribute-based-valuetypes/README.md
rhazn: I rewrote this comment three times, not sure I can come to a strong opinion so in the end I would probably accept majority opinion, but:

I think that is unfortunate because it again introduces magic with the `value` name for builtin value types. It also still means it is weird to reuse constraints between inline definitions and general definitions. I would prefer if we keep `value` in general for one-dimensional value types.

I can not really think of a good way to model constrains on multi-dimensional value types that do not need to know the names of properties... For consistencies sake I would go with `value.x`. That does not resolve the issue of course, but it keeps the rule intact that `value` always refers to the underlying value of the value type.

rfc/0014-attribute-based-valuetypes/README.md
georg-schwarz: I don't know, I just took it from a previous version :)

georg-schwarz: Reformulated it a little bit

rfc/0014-attribute-based-valuetypes/README.md
georg-schwarz: Ok, I see two alternatives:

## Alternative 1:  Inline constraints use property names

So in the example, `correlation >= -1 and correlation <= 1`?

This leads to different syntax for inline and referenceable constraints. 
However, users can call their property of single-attribute value types "value," and we can promote that through our documentation. Then, there is no syntactic difference when extracting an in-line constraint to a referenceable constraint.

```
// Referenceable constraint
valuetype CorrelationCoefficient {
  property correlation oftype decimal; 
  constraint minusOneToPlusOneRange: CorrelationCoefficient on correlation;
}
constraint CorrelationCoefficient on decimal:
  value >= -1 and value <= 1;

// Inline constraint (not following convention)
valuetype CorrelationCoefficient {
  property correlation oftype decimal; 

  constraint minusOneToPlusOneRange:
    correlation >= -1 and correlation <=1;
}

// Inline constraint (following convention)
valuetype CorrelationCoefficient {
  property value oftype decimal; // convention: should be called value for one-dim value types

  constraint minusOneToPlusOneRange:
    value >= -1 and value <=1;
}
```

## Alternative 2:  Constraints always use `value`

So in the example, `value.correlation >= -1 and value.correlation <= 1`. 
This alternative leads to similar syntax for inline and referenceable constraints.

```
// Referenceable constraint 
valuetype CorrelationCoefficient {
  property correlation oftype decimal; 
  constraint minusOneToPlusOneRange: CorrelationCoefficient on correlation;
}
constraint CorrelationCoefficient on decimal:
  value >= -1 and value <= 1;

// Inline constraint (not following convention)
valuetype CorrelationCoefficient {
  property correlation oftype decimal; 

  constraint minusOneToPlusOneRange on correlation:
    value >= -1 and value <=1;
}
```

The question remaining for this alternative  is how to model multi-dimensional constraints, like @rhazn already mentioned... I tried a few times, but didn't come to a good solution.

georg-schwarz:
@rhazn This PR seems outdated, but I remember incorporating all the changes we spoke about (as far as I was able to).
Can we merge this?

@TungstnBallon will probably take a turn to implement (a part of) the changes.

rhazn:
Yes, I'd merge this. It is definitely better and if I remember right I had no very strong opinions on the discussion either, if we can refresh it and take a short vote I am sure we can figure something good out :).

PR #598, opened 2024-07-02
Add CSVFIleLoader block

### LINKED ISSUE #394: [FEATURE] Save table as CSV
## User Story
1. As a {user/persona}
2. I want {to save data as CSV}
3. So that {I can share the CSV}

## User Acceptance Criteria
- [ ] A new block "CSVFileLoader" is available, like SQLiteLoader. It has a `Table` input,`None` output and the required property "file".
- [ ] It has all properties of the CSVInterpreter as optional properties with the following defaults: delimiter: ",", enclosing: "", enclosingEscape: "".
- [ ] Saves a CSV file with the table data at the `file` location, relative to the jv model file, with the delimiter, enclosing and enclosingEscape characters as configured, uses utf-8 as character encoding

## Notes
- https://jvalue.github.io/jayvee/docs/user/block-types/CSVInterpreter
- https://jvalue.github.io/jayvee/docs/user/block-types/SQLiteLoader
- https://github.com/jvalue/jayvee/blob/main/libs/execution/src/lib/types/valuetypes/visitors/sql-value-representation-visitor.ts

## Definitions of Done
- [ ] All tests are passing

TungstnBallon:
This PR adds a the ability to export tables into csv files using `fast-csv`, the same library that parses the csv in `CSVIntperpreter`.

closes #394 

TODO before merging:
 - [x] Remove commit 3b49022

### DIFF
--- a/libs/extensions/tabular/exec/src/lib/csv-file-loader-executor.spec.ts
+++ b/libs/extensions/tabular/exec/src/lib/csv-file-loader-executor.spec.ts
@@ -2,8 +2,8 @@
 //
 // SPDX-License-Identifier: AGPL-3.0-only
 
+import * as fsPromise from 'node:fs/promises';
 import path from 'node:path';
-import * as fsPromise from 'fs/promises';
 
 import * as R from '@jvalue/jayvee-execution';
 import {
@@ -29,8 +29,8 @@ import {
   type LangiumDocument,
 } from 'langium';
 import { NodeFileSystem } from 'langium/node';
-import { vi } from 'vitest';
 import { vol } from 'memfs';
+import { vi } from 'vitest';
 
 import { CSVFileLoaderExecutor } from './csv-file-loader-executor';
 
@@ -112,9 +112,9 @@ describe('Validation of CSVFileLoaderExecutor', () => {
     if (R.isOk(result)) {
       expect(result.right.ioType).toEqual(IOType.NONE);
       const expectedOutput = `Column1,Column2
-value 1, 20.2`;
+value 1,20.2`;
       const actualOutput = await fsPromise.readFile('test.csv');
-      expect(actualOutput).toEqual(expectedOutput);
+      expect(actualOutput.toString()).toEqual(expectedOutput);
     }
   });
 });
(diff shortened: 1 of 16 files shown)


libs/extensions/tabular/exec/src/lib/csv-file-loader-executor.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+        },
+      ],
+      1,
+    );
+    const result = await parseAndExecuteExecutor(text, inputTable);
+
+    expect(R.isErr(result)).toEqual(false);
+    if (R.isOk(result)) {
+      expect(result.right.ioType).toEqual(IOType.NONE);
+      // FIXME: Check that the resulting csv is valid
```
TungstnBallon, line 108: I don't really know to do this right (maybe with mock?), so I marked this pr as a draft.

TungstnBallon: @rhazn

rhazn: Yeah, you can go with whitebox testing (mocking the `writeCSVToPath` of the library and making sure it gets called with the right properties).

Alternatively, you could write the actual file to disk and then hash it's content to check them against a predefined hash. It will mean you need to carefully write and clean up files with every test run though. I think both approaches are fine :).

libs/extensions/tabular/exec/src/lib/csv-file-loader-executor.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+  });
+  beforeEach(() => {
+    // NOTE: The virtual filesystem is reset before each test
+    vol.reset();
+  });
+  afterEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('should diagnose no error on valid loader config', async () => {
```
rhazn, line 87: It would be nice to write a few more tests, one with boolean values, some with different properties (delimiter etc) set.

rhazn (changes requested):
Need to remove the commit you already identified (no need to commit local test files) and write 1-2 more tests but otherwise looks good to me 👍

rhazn (approved):

PR #605, opened 2024-07-26
Add link to SWC example to docs

rhazn:
Adds a reference to use cases to the docs, currently only with the SWC project. I'd keep the actual description of the use case on the shorter side and directly link to the GitHub repo as I think it is pretty well explained there.

I will of course not merge this PR before making the SWC examples repo public.

### DIFF
--- a/apps/docs/docs/user/examples/README.mdx
+++ b/apps/docs/docs/user/examples/README.mdx
@@ -1,5 +1,5 @@
 ---
-sidebar_position: 10
+sidebar_position: 11
 ---
 
 # Jayvee Examples
--- a/apps/docs/docs/user/use-cases/README.mdx
+++ b/apps/docs/docs/user/use-cases/README.mdx
@@ -4,7 +4,7 @@ sidebar_position: 12
 
 # Jayvee Use Cases
 
-Here you can find a selection of more complex projects that use (some form of) Jayvee to accomplish a real world task.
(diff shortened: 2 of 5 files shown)


apps/docs/docs/user/use-cases/README.mdx
```diff
@@ -0,0 +1,13 @@
+---
+sidebar_position: 12
+---
+
+# Jayvee Use Cases
+
+Here you can find a selection of more complex projects that use (some form of) Jayvee to accomplish a real world task.
```
georg-schwarz, line 7: How do you feel about adding a line that we welcome additions by users?

apps/docs/docs/user/examples/README.mdx
```diff
@@ -1,5 +1,5 @@
 ---
-sidebar_position: 10
+sidebar_position: 11
```
georg-schwarz, line 2: I assume that was just wrong and you corrected it

georg-schwarz (approved):

joluj (approved):

apps/docs/docs/user/examples/README.mdx
rhazn: Depends, we had two entries on sidebar position 10, so I assume they were ordered by alphabet. This switches the order and moves examples down so that use cases and examples are next to each other.

PR #606, opened 2024-07-30
Add code-lens to run pipeline

### LINKED ISSUE #587: [FEATURE] VSCode Extension: Run pipeline button

georg-schwarz:
Closes #587.

Adds a `Run (once)` code lens over each pipeline that executes the jv CLI.

image


apps/vs-code-extension/src/commands/run-jv.ts
```diff
@@ -0,0 +1,78 @@
+// SPDX-FileCopyrightText: 2023 Friedrich-Alexander-Universitat Erlangen-Nurnberg
```
rhazn, line 1: It's 2024 ;)

apps/vs-code-extension/src/commands/run-jv.ts
```diff
@@ -0,0 +1,78 @@
...
+  assert(
+    isRunJayveeCodeLensPayload(payload),
+    `Payload ${JSON.stringify(
+      payload,
+    )} is not a valid RunJayveeCodeLensPayload`,
+  );
+  const filePath = payload.filePath;
+
+  const shell = window.createTerminal('Run Jayvee');
+  shell.sendText(`jv ${filePath} --pipeline ${payload.pipelineName}`);
```
rhazn, line 34: Would put a `-d` in there, no?

apps/interpreter/src/run-action.ts
```diff
@@ -35,9 +34,9 @@ export async function runAction(
     return process.exit(ExitCode.FAILURE);
   }
 
-  const currentDir = getCurrentDir();
+  const currentDir = process.cwd();
   const workingDir = currentDir;
-  const absoluteFilePath = path.join(currentDir, filePath);
+  const filePathRelativeToCurrentDir = path.relative(currentDir, filePath);
```
rhazn, line 39: Does this not introduce some weirdness again that makes calling a jv file dependent on where the interpreter is running?

libs/language-server/src/lib/lsp/jayvee-code-lens-provider.ts
```diff
@@ -0,0 +1,81 @@
...
+  pipelineName: string;
+}
+
+export interface ShowPipeOutputCodeLensPayload {
+  pipeOutputName: string;
+  lineNumber: number;
+}
+
+export function isRunJayveeCodeLensPayload(
+  o: unknown,
```
rhazn, line 23: Naming could probably be better than `o`, no? At least something like 'payload'.

libs/language-server/src/lib/lsp/jayvee-code-lens-provider.ts
```diff
@@ -0,0 +1,81 @@
...
+  return (
+    'filePath' in o &&
+    typeof o.filePath === 'string' &&
+    'pipelineName' in o &&
+    typeof o.pipelineName === 'string'
+  );
+}
+
+export function isShowPipeOutputCodeLensPayload(
+  o: unknown,
```
rhazn, line 38: Same here.

apps/vs-code-extension/src/commands/run-jv.ts
```diff
@@ -0,0 +1,78 @@
...
+            `No matching jv versions: expected version ${requiredVersion} but found version ${installedVersion}`,
+          ),
+        );
+      }
+      return resolve();
+    });
+  });
+}
+
+function getExtensionVersion(context: ExtensionContext): string | undefined {
```
rhazn, line 63: This is really nice, could be used to other good effects like showing a warning if we ever add a versioning field to jv models.

rhazn (approved):
Not sure what the effects are of the directory logic changing in the `run-action.ts`, if you're confident in that change I'd trust your knowledge though.

Other comments are largely just style and can be ignored.

apps/interpreter/src/run-action.ts
georg-schwarz: > Not sure what the effects are of the directory logic changing in the run-action.ts, if you're confident in that change I'd trust your knowledge though.

Ah yes.. I noticed with our current version we stopped supporting absolute paths :(
The new directory logic enables absolute paths again, by mapping them to a relative path from the current dir.
I'm open to better solutions if you have some in mind :)

apps/vs-code-extension/src/commands/run-jv.ts
georg-schwarz: Mhm, depends. We could also add a second code action `Debug (once)`?
But in general, we should offer an option to debug, I agree!

georg-schwarz: Added in [3a61993](https://github.com/jvalue/jayvee/pull/606/commits/3a61993b7ceb13680eec75d64aacca3ea2ea00b3)

apps/interpreter/src/run-action.ts
rhazn: I think at all times, all paths should be relative to the model location without exception :).

georg-schwarz: Ah I mean for the interpreter, sorry!
So `jv /home/me/model.jv` wasn't working.
Import paths in models are always relative.

rhazn: Ah, yeah, then I love it.

PR #612, opened 2024-07-31
Add dts files to build

### LINKED ISSUE #607: [BUG] `@jvalue/jayvee-interpreter-lib` exports are broken
## Steps to reproduce
1. Write code using `@jvalue/jayvee-interpreter-lib:0.0.4`
2. Update to 0.0.5 or 0.0.6
3. Imports are broken

## Description
- Expected: The same exports are there
- Actual: Nothing is exported

---

Old (working) build output:
image

Current build output:
image

joluj:
Closes #607.

Changes:
- Update to nx 19.5.4
- All apps/libs built with `esbuild` now emit declaration files
  - These are mapped properly in the corresponding `package.json`s
- Added new publishable package `@jvalue/jayvee-execution`. This is necessary for backward compatibility since this was accessible through `@jvalue/jayvee-interpreter-lib/execution` before.
- Excluded publishable packages from library builds
  - Publishable apps must explicitly disable this in the `project.json`
  - Decided on opt-out for apps to encourage library creation

### DIFF
--- a/libs/execution/project.json
+++ b/libs/execution/project.json
@@ -16,34 +16,6 @@
         "configFile": "{projectRoot}/vite.config.ts",
         "passWithNoTests": false
       }
-    },
-    "pre-publish": {
-      "executor": "nx:run-commands",
-      "dependsOn": ["build"],
-      "options": {
-        "commands": [
-          "node tools/scripts/relax-peer-dependency-versions.mjs execution",
-          "node tools/scripts/add-package-json-version.mjs execution",
-          "node tools/scripts/publish.mjs execution false"
-        ],
-        "parallel": false
-      }
-    },
-    "publish": {
-      "executor": "nx:run-commands",
-      "dependsOn": ["pre-publish"],
-      "options": {
-        "commands": ["node tools/scripts/publish.mjs execution true"],
-        "parallel": false
-      }
-    },
-    "pack": {
-      "executor": "nx:run-commands",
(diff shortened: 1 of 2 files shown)


rhazn (approved):
I am not sure I totally follow what `Excluded publishable packages from library builds [...] Decided on opt-out for apps to encourage library creation` means so it would be great to include a small explanation in this PR for me 😅 . But otherwise the code looks fine to me 👍

joluj:
> Excluded publishable packages from library builds

If you don't exclude `@jvalue/jayvee-language-server` from `@jvalue/jayvee-interpreter-lib`, it's bundled in `@jvalue/jayvee-interpreter-lib`.

Case non-excluded (i.e. before): Assume `@jvalue/jayvee-interpreter-lib` uses class `A` from `@jvalue/jayvee-language-server`. In a third-party library, someone imports `A` from `@jvalue/jayvee-language-server`. These `A`s now look alike, but `instanceof` uses of objects of each `A` will fail because the class definition is separate.

Case excluded (i.e. after this PR): Both `A`s reference the one from `@jvalue/jayvee-language-server` :tada: 

> Decided on opt-out for apps to encourage library creation

`@jvalue/jayvee-interpreter-lib` is excluded in every built by default. If you don't want to exclude it (i.e. in the apps, we bundle everything there), you have to remove that from the exclusion list. This works by setting the option `excluded: []` in the `projects.json`.

PR #629, opened 2024-10-29
Add country code valuetypes to stdlib

georg-schwarz:


libs/language-server/src/stdlib/domain/geo/ContryCode.jv
```diff
@@ -0,0 +1,18 @@
...
+use * from "./CountryCodeNumeric.jv";
+
+/**
+* Any code (alppha2, alpha3, numeric) for countries as defined by ISO 3166.
+*/
+publish valuetype CountryCode oftype text {
+  constraints: [
+    CountryCodeAlpha2Constraint,
+    CountryCodeAlpha3Constraint,
+    CountryCodeNumericConstraint
```
georg-schwarz, line 16: Ah that might be a mistake because it might join them with an "AND"

rhazn: Indeed. I think this needs to be removed.

rhazn (approved):
Except the generic countrycode, this looks good.

PR #632, opened 2024-12-20
Autogenerate docs for composite block types

rhazn:
This PR adds autogenerated documentation for composite block types.

Because we have a lot of composite block types in domain extensions in the stdlib this PR also introduces an additional level in the documentation, blocks are now categorized as either "builtin" or by domain:
Screenshot 2024-12-20 at 3 23 34 PM

Inside each domain, the expected block definitions can be found:
Screenshot 2024-12-20 at 3 24 31 PM

### DIFF
--- a/apps/docs-generator/src/UserDocCategoryBuilder.ts
+++ b/apps/docs-generator/src/UserDocCategoryBuilder.ts
@@ -1,0 +1,66 @@
+// SPDX-FileCopyrightText: 2024 Friedrich-Alexander-Universitat Erlangen-Nurnberg
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+
+import * as fs from 'node:fs';
+import path from 'node:path';
+
+export class UserDocCategoryBuilder {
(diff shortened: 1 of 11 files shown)


rhazn:
Not sure I understand why the license compliance fails. The error seems to be 

```
reuse._util - ERROR - Could not parse 'AGPL-3.0-only`;'
reuse.project - ERROR - 'apps/docs-generator/src/UserDocCategoryBuilder.ts' holds an SPDX expression that cannot be parsed, skipping the file
```

But I can spot no mistake in the license definition there? Does anyone else have a hunch?

georg-schwarz (approved):
Awesome :partying_face: !

Does this also list the composite block types on root level of the stdlib, like the [CSVExtractor](https://github.com/jvalue/jayvee/blob/main/libs/language-server/src/stdlib/CSVExtractor.jv)?

rhazn:
It does, they get listed as `Built-in Blocktypes` (imho correctly) :).

PR #635, opened 2025-01-10
[FEATURE] Add custom hook support

### LINKED ISSUE #633: [FEATURE] Add custom hook support
## User Story
1. As a {developer that uses the Jayvee lib}
2. I want {to add custom hooks that run before or after blocks}
3. So that {I can enrich Jayvee without modifying the original source code}

## User Acceptance Criteria
- [ ] Users of the Jayvee lib can add hooks to all blocks
- ~Users of the Jayvee lib can add hooks to individual blocks~ (not in the first version)
- ~Users of the Jayvee lib can add hooks to whole pipelines / programs~ (not in the first version)
- [ ] The number of hooks per blocks is not limited

A hook is basically a function that is called when due.
- [ ] It can be blocking or non-blocking. It's a user's choice. Default: non-blocking
- [ ] It can be asynchronous (important detail for blocking functions).
- [ ] If it's a pre-block hook: It retrieves the input parameters of a block (i.e. metadata and data).
- [ ] If it's a post-block hook: It retrieves the input and output parameters of a block (i.e. metadata and input data and output data).
- [ ] In this first version, there is no option to alter the input or output data or metadata. 

## Examples

I can imaging that we can use it like that:

```ts
const myJayveeProgram = // ...;
myJayveeProgram.addPreBlockHook(async ({args, input}) => {
  console.log(args); // type `unknown`? For now at least
  console.log(input); // type `unknown`? For now at least
}, {
  blocking: true,
});

myJayveeProgram.execute();
```

## Notes

No support is needed for:
- Removing hooks
- 

## Definitions of Done
- [ ] All tests are passing

### LINKED ISSUE #634: [FEATURE] Add custom hook support for specific blocks
This is a follow-up issue of #633.

## User Story
1. As a {developer that uses the Jayvee lib}
2. I want {to add custom hooks that run before or after blocks}
3. So that {I can enrich Jayvee without modifying the original source code}

## User Acceptance Criteria
- [ ] I can add custom hooks to specific block types via reference to the class
- [ ] I can add custom hooks to specific block types via a boolean function with the param `{ type: string }` (or slightly different if it fits better to the current implementation)

- [ ] The arguments (meta, input, output data) for the hook functions are typed properly.
  - This may be disabled for hooks with a `blocks` array of size >1. Here, the types may remain `unknown`.

- [ ] The implementation should be backwards compatible, i.e. the `blocks` argument should be optional.

## Examples

```ts
import { HttpExtractor } from '...';

myJayveeProgram.addPreBlockHook(myLambda, {
  blocks: [HttpExtractor],
})

// Or for custom blocks like composite blocks (they are not built-in)
myJayveeProgram.addPreBlockHook(myLambda, {
  blocks: [HttpExtractor, ({ type }) => type === 'MyCustomCompositeBlock'],
})
```

## Notes

- Try to keep it as simple as possible in regards to the implementation/design (which probably means it's not efficient as possible)
  - E.g., Map a missing list of `blocks` to `[() => true]`
  - E.g., Map a class that was given to the `blocks` array to a function internally (`HttpExtractor` => `({ type }) => type === HttpExtractor.type`).

## Definitions of Done
- [ ] All tests are passing

TungstnBallon:
Closes #633.

This PR adds hook  functionality to the Jayvee interpreter. Hooks are custom functions that are automatically executed before / after blocks. See the `PreBlockHook` and `PostBlockHook` types for such a functions signature.

Library users can add hooks with the `addHook` method:
```typescript
const program = /* parse program ...*/;

const before: PreBlockHook = (blocktype, input, context) => { /* ... */ };
program.addHook('preBlock', before, { blocking: true });

const after: PostBlockHook = (blocktype, input, context, output) => { /* ... */ };
program.addHook('postBlock', after);  // Options can be omitted. Non-blocking is the default

const blockSpecificHook = (blocktype, input, context) => { /* ... */ };
myJayveeProgram.addHook('preBlock', blockSpecificHook, {
  blocktypes: ['BuiltinBlockType', 'MyCustomCompositeBlock'],
})

// interpret model ...
```

Edit: Due to the simplifications outlined in https://github.com/jvalue/jayvee/issues/634#issuecomment-2589640027 this PR also closes #634.

### DIFF
--- a/libs/execution/src/lib/hooks/hook-context.ts
+++ b/libs/execution/src/lib/hooks/hook-context.ts
@@ -150,11 +150,15 @@ export class HookContext {
     output?: Result<IOTypeImplementation | null>,
   ) {
     if (output === undefined) {
+      context.logger.logInfo(`Executing general pre-block-hooks`);
       const general = executeTheseHooks(
         this.hooks.pre[AllBlocks] ?? [],
         blocktype,
         input,
         context,
+      );
+      context.logger.logInfo(
+        `Executing pre-block-hooks for blocktype ${blocktype}`,
       );
       const blockSpecific = executeTheseHooks(
         this.hooks.pre[blocktype] ?? [],
@@ -166,12 +170,16 @@ export class HookContext {
       // eslint-disable-next-line @typescript-eslint/no-empty-function
       return Promise.all([general, blockSpecific]).then(() => {});
     }
+    context.logger.logInfo(`Executing general post-block-hooks`);
     const general = executeTheseHooks(
       this.hooks.post[AllBlocks] ?? [],
       blocktype,
       input,
       context,
       output,
+    );
+    context.logger.logInfo(
+      `Executing post-block-hooks for blocktype ${blocktype}`,
     );
     const blockSpecific = executeTheseHooks(
       this.hooks.post[blocktype] ?? [],
--- a/libs/execution/src/lib/hooks/hook-context.ts
+++ b/libs/execution/src/lib/hooks/hook-context.ts
@@ -85,12 +85,7 @@ export class HookContext {
     hook: PreBlockHook | PostBlockHook,
     opts: HookOptions,
   ) {
-    const blocktypes: string[] =
-      typeof opts.blocktypes === 'string'
-        ? [opts.blocktypes]
-        : opts.blocktypes ?? [AllBlocks];
-
-    blocktypes.forEach((blocktype) => {
+    for (const blocktype of opts.blocktypes ?? [AllBlocks]) {
       if (isPreBlockHook(hook, position)) {
         if (this.hooks.pre[blocktype] === undefined) {
           this.hooks.pre[blocktype] = [];
@@ -108,7 +103,7 @@ export class HookContext {
           hook,
         });
       }
-    });
+    }
   }
 
   public async executePreBlockHooks(
--- a/libs/execution/src/lib/hooks/hook-context.ts
+++ b/libs/execution/src/lib/hooks/hook-context.ts
@@ -1,3 +1,7 @@
+// SPDX-FileCopyrightText: 2025 Friedrich-Alexander-Universitat Erlangen-Nurnberg
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+
 // eslint-disable-next-line unicorn/prefer-node-protocol
 import assert from 'assert';
 
--- a/libs/execution/src/lib/hooks/hook-context.ts
+++ b/libs/execution/src/lib/hooks/hook-context.ts
@@ -1,0 +1,183 @@
+// eslint-disable-next-line unicorn/prefer-node-protocol
+import assert from 'assert';
+
+import { type Result } from '../blocks';
+import { type ExecutionContext } from '../execution-context';
+import { type IOTypeImplementation } from '../types';
+
+import {
+  type HookOptions,
+  type HookPosition,
+  type PostBlockHook,
+  type PreBlockHook,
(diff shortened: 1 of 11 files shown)


libs/interpreter-lib/src/interpreter.spec.ts
```diff
@@ -26,4 +27,141 @@ describe('Interpreter', () => {
       expect(exitCode).toEqual(ExitCode.SUCCESS);
     });
   });
+
+  describe('hooks', () => {
```
joluj, line 31: Can you add a test to check if the input/output is called with the correct value?

libs/execution/src/lib/hooks/hook.ts
```diff
@@ -0,0 +1,46 @@
...
+import { type IOTypeImplementation } from '../types';
+
+/** When to execute the hook.*/
+export type HookPosition = 'preBlock' | 'postBlock';
+
+export interface HookOptions {
+  /** Whether the pipeline should await the hooks completion. `false` if omitted.*/
+  blocking?: boolean;
+  /** Optionally specify one or more blocks to limit this hook to. If omitted, the hook will be executed on all blocks*/
+  blocktypes?: string | string[];
```
joluj, line 16: `string[]` instead of `string | string[]` would be fine here I think. The attribute's name is plural, which is weird if it's a `string` I feel like.

libs/execution/src/lib/hooks/hook-context.ts
```diff
@@ -0,0 +1,195 @@
...
+} from './hook';
+
+const AllBlocks = '*';
+
+interface HookSpec<H extends PreBlockHook | PostBlockHook> {
+  blocking: boolean;
+  hook: H;
+}
+
+async function executeTheseHooks(
```
joluj, line 27: I think it's better so split these functions into two. The typing is quite complex and will get more complex the more functionality we add. 
When we add the functionality to alter the input/output value, the logic will be too different as well I think.

joluj (approved):
Nice work!

I think that the logic will get quite complicated in the future with the current setup. But we can fix it as soon as it is a problem.

I'm curious why the pre and post hooks are not simply added in `AbstractBlockExecutor::execute`?

libs/execution/src/lib/hooks/hook-context.ts
```diff
@@ -0,0 +1,195 @@
...
+  public addHook(
+    position: 'postBlock',
+    hook: PostBlockHook,
+    opts: HookOptions,
+  ): void;
+  public addHook(
+    position: HookPosition,
+    hook: PreBlockHook | PostBlockHook,
+    opts: HookOptions,
+  ): void;
```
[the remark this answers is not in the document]
TungstnBallon: I decided to split this method into a preBlock and postBlock variant too.

TungstnBallon:
> I'm curious why the pre and post hooks are not simply added in `AbstractBlockExecutor::execute`?

I didn't want hooks to affect the block duration measured `executeBlock`:
https://github.com/jvalue/jayvee/blob/202983628666f26ed612693bfca40b9b9672503a/libs/execution/src/lib/blocks/block-execution-util.ts#L114

libs/execution/src/lib/hooks/hook-context.ts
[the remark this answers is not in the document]
joluj: I think you can omit lines 78-82 because of lines 83-87 but I'm not sure. No need for change here, I was just unsure about the TS syntax.

joluj (approved):
Awesome! 🎉

PR #638, opened 2025-01-15
[FEATURE] Include hooks in time measurement of blocks

TungstnBallon:
Closes #637.

This PR adds the hook duration to the debug logger. This is the same level as block duration logging.

PR #639, opened 2025-01-21
[REFRACTOR] Hooks recieve a single parameter

TungstnBallon:
This PR changes the hook's signature to only receive one object with the previous parameter as properties. This has the following advantages:
- Backwards compatibility: Adding new "parameters" at whatever position doesn't break existing code
- It's easier to ignore unused "parameters".

Depends on #638.

### DIFF
--- a/libs/execution/src/lib/execution-context.ts
+++ b/libs/execution/src/lib/execution-context.ts
@@ -155,15 +155,20 @@ export class ExecutionContext {
     );
 
     if (output === undefined) {
-      return this.hookContext.executePreBlockHooks(blocktype, input, this);
+      return this.hookContext.executePreBlockHooks({
+        blocktype,
+        input,
+        context: this,
+      });
+      // eslint-disable-next-line no-else-return
+    } else {
+      return this.hookContext.executePostBlockHooks({
+        blocktype,
+        input,
+        output,
+        context: this,
+      });
     }
-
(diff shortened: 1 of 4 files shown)


libs/execution/src/lib/execution-context.ts
```diff
@@ -155,15 +155,20 @@ export class ExecutionContext {
     );
 
     if (output === undefined) {
-      return this.hookContext.executePreBlockHooks(blocktype, input, this);
+      return this.hookContext.executePreBlockHooks({
+        blocktype,
+        input,
+        context: this,
+      });
+      // eslint-disable-next-line no-else-return
```
joluj, line 163: Why disable and not apply the rule?

joluj (approved):
:tada:

libs/execution/src/lib/execution-context.ts
TungstnBallon: For me, code like
```typescript
if (cond) {
  return a;
}

return b;
```
communicates that `b` is the usual result and `a` an edge case, because `a` is at a higher indentation level than `b`.

Here however, `executePreBlockHook` is as valid as `executePostBlockHook`, so to me, it makes sense to have them at the same indentation level.

This is just personal preference though, If the former fits the project's style better I can change it.

joluj: That totally makes sense! Thanks

PR #644, opened 2025-01-29
[PERF] Don't split and join lines when using `TextFile`

TungstnBallon:
**Problem**: When chaining `TextRangeSelector` and `TextLineDeleter` the following happens:
```typescript
// TextRangeSelector
let lines = splitLines(file.content);
let newLines = /* ... select range ... */;
return new TextFile(newLines.join());

// ...

// TextLineDeleter
let lines = splitLines(file.content);
let newLines = /* ... deleteLines ... */;
return new TextFile(newLines.join());
```
This might mean a lot of "useless" splits and joins, which becomes especially problematic for large files.

**Solution**: Introduce the new util function `findLineBounds()`. This function takes a sorted array of line indices, a regex to find line breaks and the text as parameters. It maps each line index to the start of the line and its length, stopping at the first non existent line index.

I didn't remove the now unused `transformTextFileLines()`, because deleting working, tested code seems unnecessary, but let me know it you think otherwise.

### DIFF
--- a/libs/execution/src/lib/debugging/debug-log-visitor.ts
+++ b/libs/execution/src/lib/debugging/debug-log-visitor.ts
@@ -6,6 +6,7 @@ import {
   type WrapperFactoryProvider,
   internalValueToString,
 } from '@jvalue/jayvee-language-server';
+import { either } from 'fp-ts';
 
 import { type Logger } from '../logging/logger';
 import { type Workbook } from '../types';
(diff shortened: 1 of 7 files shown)


libs/execution/src/lib/debugging/debug-log-visitor.ts
```diff
@@ -115,18 +117,31 @@ export class DebugLogVisitor implements IoTypeVisitor<void> {
...
+    switch (this.debugGranularity) {
+      case 'minimal':
+        return;
+
+      case 'exhaustive':
+        this.log(textFile.content);
+        return;
+
+      case 'peek': {
+        // BUG: /\r?\n/ might not be the correct line break
```
joluj, line 130: Can you create a Bug issue for that right away?

joluj (approved):

libs/execution/src/lib/debugging/debug-log-visitor.ts
TungstnBallon: Done (#645)

Issue #641, opened 2025-01-27
[BUG] Cannot read properties of undefined (reading 'length')


ozalieg:
## Steps to reproduce
1. Just run the jv pipeline I added at the end over terminal (jv \<filename\>).

## Description
- Expected: no problems
- Actual:
```
        error: An unknown error occurred: TypeError: Cannot read properties of undefined (reading 'length')
$In /Users/lea/Documents/Uni/WS24_25/Thesis/ws2425-buchner-jayvee-templating/General/TemplateGeneration/JvFiles/BiostatsPipeline.jv:16:11
        16 |     block BiostatsCSVInterpreter oftype CSVInterpreter {
           |    
```

## Example Pipeline where the error occurs
```
pipeline BiostatsPipeline {

    BiostatsExtractor
    -> BiostatsTextFileInterpreter
    -> BiostatsCSVInterpreter
    -> BiostatsTableInterpreter
    -> BiostatsLoader;

    block BiostatsExtractor oftype HttpExtractor {
        url: "https://people.sc.fsu.edu/~jburkardt/data/csv/biostats.csv";
    }

    block BiostatsTextFileInterpreter oftype TextFileInterpreter {
    }

    block BiostatsCSVInterpreter oftype CSVInterpreter {
        enclosing: "";
    }

    block BiostatsTableInterpreter oftype TableInterpreter {
        header: true;
        columns: [
                "Name" oftype text,
                "Sex" oftype text,
                "Age" oftype integer,
                "Height (in)" oftype integer,
                "Weight (lbs)" oftype integer
        ];
    }

    block BiostatsLoader oftype SQLiteLoader {
        table: "Biostats";
        file: "./Biostats.sqlite";
    }

}
```

TungstnBallon:
The above `TypeError` will be fixed by #642.
However, the pipeline still doesn't work as intended, because the ` "Age"` column contains numbers with a leading whitespace. 
Jayvee (currently) doesn't consider these values numbers: https://github.com/jvalue/jayvee/blob/186dcb722fb56fdbb0fdd26ae5aaf0c877c5aa5b/libs/language-server/src/lib/ast/wrappers/value-type/primitive/decimal-value-type.ts#L10


@joluj What would you say is the intended behavior here?

joluj:
AFAIK leading whitespaces are part of the cell in CSV. Therefore it's the problem with the CSV. 

[...]

PR #647, opened 2025-02-04
[BUG] Enable `TableInterpreter` to trim leading and trailing whitespace

TungstnBallon:
Adds two properties to `TableInterpreter`: `skipLeadingWhitespace` and `skipTrailingWhitespace`. Both are true by default.
If enabled `InternalRepresentationParserVisitor` will trim whitespace for `integer`, `decimal` and `boolean`.

Part of #641.


georg-schwarz (approved):
LGTM.

Having it in the `TableInterpreter` seems a good solution to me.

We might want to define it column-specific in the future, esp. if we have a nicer syntax to apply transformations. Example (just spitballing):
```
    block TestBlock oftype TableInterpreter {
		header: false;
		columns: [
			"index" oftype integer <- stripWhitespaces,
			"name" oftype text,
			"flag" oftype boolean <- stripWhitespaces
		];\
	}
```

libs/execution/src/lib/types/value-types/internal-representation-parsing.ts
```diff
@@ -18,31 +18,59 @@ import {
...
->(value: string, valueType: ValueType<I>): I | undefined {
-  const visitor = new InternalRepresentationParserVisitor(value);
+>(
+  value: string,
+  valueType: ValueType<I>,
+  parseOpts?: ParseOpts,
+): I | undefined {
+  const visitor = new InternalRepresentationParserVisitor(
+    value,
+    parseOpts ?? { skipLeadingWhitespace: true, skipTrailingWhitespace: true },
```
georg-schwarz, line 28: Consider pulling the default out to a `defaultParsOpts` object

libs/execution/src/lib/types/value-types/internal-representation-parsing.ts
```diff
@@ -18,31 +18,59 @@ import {
...
+
 class InternalRepresentationParserVisitor extends ValueTypeVisitor<
   InternalValueRepresentation | undefined
 > {
-  constructor(private value: string) {
+  constructor(private value: string, private parseOpts: ParseOpts) {
     super();
   }
 
+  private trim() {
```
georg-schwarz, line 49: Suggestion for alternative name: `applyTrimOptions`

libs/execution/src/lib/types/value-types/internal-representation-parsing.ts
```diff
@@ -18,31 +18,59 @@ import {
...
+    if (this.parseOpts.skipLeadingWhitespace) {
+      this.value = this.value.trimStart();
+    }
+    if (this.parseOpts.skipTrailingWhitespace) {
+      this.value = this.value.trimEnd();
+    }
+  }
+
   visitBoolean(vt: BooleanValuetype): boolean | undefined {
+    this.trim();
```
georg-schwarz, line 63: Not sure whether manipulating the `value` is the way to go vs. just handing through the manipulated value to the next transformation. What was your rationale here to do it like that?

joluj (approved):
I agree with @georg-schwarz's comments. 

> We might want to define it column-specific in the future

I agree, but it's something for another PR.

joluj (approved):

libs/execution/src/lib/types/value-types/internal-representation-parsing.ts
TungstnBallon: This is just what occurred to me first. I do agree though, mutating `this.value` here isn't really necessary.

Issue #637, opened 2025-01-15
[FEATURE] Include hooks in time measurement of blocks


joluj:
## User Story
1. As a {developer}
2. I want {to see how my hooks influence the performance of blocks}
3. So that {I can improve the execution speed}

## User Acceptance Criteria
- [ ] Time tracking still works similar to now
- [ ] Time tracking includes information about time spent in pre-block hooks, in the block itself, and in post-block hooks.
- [ ] In the default logging mode, the full time for the block including all (blocking) hooks should be logged. In `debug` logging mode, the individual times should also be logged. 

## Notes

On which log level do we currently log the execution time?

## Definitions of Done
- [ ] A PR has been opened and accepted
- [ ] All user acceptance criteria are met
- [ ] All tests are passing

Issue #636, opened 2025-01-15
[FEATURE] Move block-related logging to a hook


joluj:
## User Story
1. As a {developer}
2. I want {have as simple code as possible}
3. So that {I can read the code}

## User Acceptance Criteria
- [ ] Evaluated if it's actually simpler if the block-related generic logging is implemented as a hook.
- [ ] (If so:) Implement the logging as a hook.

## Examples

/

## Notes

Please evaluate it first if it's a good addition or not.

## Definitions of Done
- [ ] A PR has been opened and accepted
- [ ] All user acceptance criteria are met
- [ ] All tests are passing

joluj:
Okay, thanks for the explanation. Then I'd suggestion closing this issue.

PR #649, opened 2025-02-20
Replace `Date` with `perf-hooks` for time measurement

TungstnBallon:
Currently, Jayvee uses [Date](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date)  to measure the duration of hooks, blocks and pipelines.

With this PR, the Jayvee interpreter now sets ["performance marks"](https://nodejs.org/api/perf_hooks.html) at relevant points during execution (before/after blocks/pipelines) and then measures the time between these marks. For now, marks are named with the following schema, with the goal of "encoding" the mark's location in the name:
```
`${pipeline name}::${block name}::${preBlockHook or BlockExecution or postBlockHook}::{start or end}`
```
This isn't final though, and I other ideas are welcome.

IMO this is a better approach than using the standard Jayvee hooks. Quoting https://github.com/jvalue/jayvee/issues/636#issuecomment-2593209146:
> As of now, the hooks' execution order is not sophisticated enough to log execution duration. Currently generic hooks are executed before block specific hooks in the order they are added. Considering https://github.com/jvalue/jayvee/issues/637, the hook logging execution duration would have to be executed after all other hooks, which cannot be guaranteed at this point.

I hope, that performance hooks can be used to implement a benchmarking tool using [PerformanceObserver](https://nodejs.org/api/perf_hooks.html#class-performanceobserver). That need's further testing though, so this PR is a draft for now.

### DIFF
--- a/libs/execution/src/lib/blocks/block-execution-util.ts
+++ b/libs/execution/src/lib/blocks/block-execution-util.ts
@@ -98,69 +98,63 @@ export async function executeBlock(
     name: block.name,
     type: blockType,
   });
-  const { result, durationMs: totalDurMs } = await measure(
-    location,
-    async () => {
-      assert(location.block !== undefined);
+  const { result, durationMs: totalDurMs } = await measure(async () => {
+    assert(location.block !== undefined);
 
-      const { durationMs: preDurMs } = await measure(
-        location.withBlockInternalLocation('preBlockHooks'),
-        () => executionContext.executeHooks(inputValue),
-      );
-      executionContext.logger.logDebug(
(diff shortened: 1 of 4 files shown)


joluj (approved):
Looks good! What's still missing?

TungstnBallon:
> Looks good! What's still missing?

I wanted to verify that this can actually be used to create a benchmarking tool (see #650).

joluj (approved):
Nice work! 

Could you add a bit more documentation, especially to functions exported from the lib.

joluj (approved):
Awesome thank you!

PR #650, opened 2025-02-20
Add benchmarking into the CI

TungstnBallon:
Depends on #649.

This PR implements a simple benchmarking tool (`npm run benchmark`) and adds a CI step to verify that no performance regression is introduced.

Benchmarks are specified in `apps/benchmark/src/benchmark_definitions.json`. For now, only the total pipeline duration is tested.

Failed benchmark step: https://github.com/jvalue/jayvee/actions/runs/14313480465/job/40113976972

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -45,6 +45,9 @@ jobs:
       - name: Test
         run: npx nx affected --target=test --verbose --parallel=1 --base=remotes/origin/main --head=HEAD --ci
 
+      - name: Benchmark
+        run: npm run benchmark
+
       # Ensure this works for future release publishing
       - name: Publish Dry-Run
         run: npx nx affected --target=pre-publish --verbose --parallel=1 --base=remotes/origin/main --head=HEAD --ci
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -45,9 +45,6 @@ jobs:
       - name: Test
         run: npx nx affected --target=test --verbose --parallel=1 --base=remotes/origin/main --head=HEAD --ci
 
-      - name: Benchmark
-        run: npm run benchmark
-
       # Ensure this works for future release publishing
       - name: Publish Dry-Run
         run: npx nx affected --target=pre-publish --verbose --parallel=1 --base=remotes/origin/main --head=HEAD --ci
@@ -55,3 +52,23 @@ jobs:
       # Ensure this works for future release publishing
       - name: Pack VSCode Extension
         run: npx nx run vs-code-extension:pack:prod
+
+  benchmark:
+    name: benchmark
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 'lts/*'
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Benchmark
+        run: npm run benchmark
+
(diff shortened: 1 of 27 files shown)


.github/workflows/ci.yaml
```diff
@@ -45,6 +45,9 @@ jobs:
       - name: Test
         run: npx nx affected --target=test --verbose --parallel=1 --base=remotes/origin/main --head=HEAD --ci
 
+      - name: Benchmark
+        run: npm run benchmark
+
```
joluj, line 50: Can you move that to another CI job called `benchmark` or so in this file? I assume that we extend the functionality of this benchmark system soon and then we don't clutter the normal CI job with it.

apps/benchmark/src/index.ts
```diff
@@ -0,0 +1,65 @@
...
+      break;
+    }
+    case '>': {
+      process.exitCode = 2;
+      break;
+    }
+    case '=': {
+      throw new Error("`cmp` cannot be `'='`");
+    }
+  }
```
joluj, line 62: This means that the CI also fails when the new system is too fast right?

apps/benchmark/src/index.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  const outOfBounds = results.find(({ cmp }) => cmp !== '=');
+
+  if (outOfBounds === undefined) {
+    console.info('No anomalies in the benchmark');
+    process.exitCode = 0;
+    return;
+  }
+
+  console.warn(outOfBounds);
```
joluj, line 48: This only warns for the first anomaly right? So if there are 10 out of 20 benchmarks failing, the CI only prints one warning?

apps/benchmark/src/index.ts
```diff
@@ -0,0 +1,65 @@
...
+import { type PipelineMeasure } from '@jvalue/jayvee-execution';
+
+import benchmarkDefinitions from './benchmark_definitions.json';
+import { createInterpreter, runBenchmark } from './run';
+
+function measureCompare(
+  actual: PipelineMeasure,
+  expected: PipelineMeasure,
+  allowedDeviationFactor: number,
+): '=' | '<' | '>' {
```
joluj, line 14: I'd use `tooSlow` instead of `<` and `tooFast` instead of `>`. Makes it a bit easier to read is the calling code.

joluj (changes requested):
Nice work :partying_face: 

I found 3 things that need change (`MeasureMENT`, CI, and more logging), but I assume that all of it is easy to fix. The rest is optional.

apps/benchmark/src/index.ts
TungstnBallon: Yes, because IMO, if we introduce a performance optimization, the expected values in `benchmark_definitions.json` should also have to be edited to reflect that optimization.

apps/benchmark/src/calc.ts
```diff
@@ -0,0 +1,66 @@
...
+
+function avgWith(
+  oldAverage: number,
+  newValue: number,
+  newSize: number,
+): number {
+  return oldAverage + (newValue - oldAverage) / newSize;
+}
+
+function avgBlkMeasure(b1: BlockMeasure, b2: BlockMeasure, bIdx: number) {
```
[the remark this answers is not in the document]
joluj: I would still go for `Measurement`. 

From what I've read and from ChatGPT:
`Measurement` implies the result of the process of measuring something.
`Measure` typically refers to the specific unit, method, or process of measuring. 

I think `Measure` kinda fits, but `Measurement` fits better as I understand it as the outcome of measuring something. That's also what I know from literature, so I'd go with that.

joluj (approved):
Thank you!

PR #651, opened 2025-03-11
Bump actions versions

joluj:


rhazn (approved):
I quickly double checked and the with -> path property of uploading seems to still exist in v3 so as far as I know this looks promising ;).

PR #663, opened 2025-05-23
[REFACTORING] Remove builtin constraint types

### LINKED ISSUE #662: [REFACTORING] Remove builtin constraint types
Can we delete the builtin constraint types? @rhazn 
Removing it might make implementing RFC 0014 easier.

I think this might be a legacy thing that we only created until we had constraint expressions.
For domain-specific constraint types we could also implement operators like `isXYDomainValidBinaryThing` or so.

TungstnBallon:
Closes #662.

This PR removes builtin constraint types from the codebase, making expression constraints the only type of constraint.


libs/interpreter-lib/src/validation-checks/runtime-parameter-literal.ts
```diff
@@ -65,7 +61,7 @@ function checkRuntimeParameterValueParsing(
     return;
   }
 
-  const propertySpec = wrapper?.getPropertySpecification(propertyName);
+  const propertySpec = wrapper.getPropertySpecification(propertyName);
```
georg-schwarz, line 64: Are you sure wrapper cannot be undefined here?

georg-schwarz (approved):
That was fast :smile: Good job!
I think we can merge this.

From removing the many tests I questioned myself whether the expression constraints are adequately tested in its current state. However, rather a topic for a follow-up issue if you feel the same!

TungstnBallon:
I looked into it and it seems there is:
- 1 test for a valid constraint.
- 1 test for a constraint with an incompatible type
- 1 test where the constraint's expression can be simplified

I agree that it's better to handle this in a follow up though.

PR #664, opened 2025-05-23
[RFC14] Part 1: Replace valuetype inheritance with one attribute

TungstnBallon:
Part 1 of implementing RFC14.

This PR replaces the valuetype's existing inheritance syntax with composition, specifically one attribute.

Example of the previous syntax:
```
valuetype SomeName oftype text {
    constraints: [
        // ...
    ];
}
```

Example of the new syntax:
```
valuetype SomeName {
    property someAttributeName oftype text;
    constraints: [
        // ...
    ];
}
```
For now, there has to be exactly one attribute, the name of which is completely irrelevant.

### DIFF
--- a/libs/language-server/src/lib/ast/wrappers/value-type/atomic-value-type.ts
+++ b/libs/language-server/src/lib/ast/wrappers/value-type/atomic-value-type.ts
@@ -90,8 +90,7 @@ export class AtomicValueType
   }
 
   protected override doGetSupertype(): ValueType | undefined {
-    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
-    const supertype = this.astNode?.type;
+    const supertype = this.astNode.attribute?.type;
     return this.wrapperFactories.ValueType.wrap(supertype);
   }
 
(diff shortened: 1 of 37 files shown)


libs/language-server/src/lib/ast/wrappers/value-type/atomic-value-type.ts
```diff
@@ -90,8 +90,7 @@ export class AtomicValueType
   }
 
   protected override doGetSupertype(): ValueType | undefined {
-    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
-    const supertype = this.astNode?.type;
+    const supertype = this.astNode.attribute?.type;
```
georg-schwarz, line 93: Are you sure we can assume that this.astNode is always defined?

georg-schwarz (approved):
Awesome keeping the scope so narrow, was a breeze to review! :partying_face: 

Left some nitpicky comments, feel free to ignore the one about test item naming.

libs/language-server/src/lib/ast/wrappers/value-type/atomic-value-type.ts
TungstnBallon: I forgot disabling the rule and eslint automatically removed the question mark

georg-schwarz (approved):

PR #666, opened 2025-06-05
Replace length access with operator

TungstnBallon:
This PR replaces the existing syntax
```
constraint ExactlyTenCharacters on text: value.length == 10;
```
with
```
constraint ExactlyTenCharacters on text: lengthof value == 10;
```
This allows lengths of any string, collection or reference to either to be calculated, not just `value`.


georg-schwarz (approved):
LGTM! Nicely done 🎉

PR #668, opened 2025-07-03
[FEATURE] Validate that constraint expressions contain the correct free variable 

TungstnBallon:
This PR adds the following validation to constraint expressions:
- **Inline constraint expressions**: must contain at least one reference to the valuetype's property
- **Constraint Definitions**: must contain at least one `ValueKeywordLiteral`

Additionally, `toHaveBeenCalledWith` is replaced with `toHaveLastBeenCalledWith`, see cdc2622677874f58b3226997ddeb1d27ae6a4ad7's commit message for more information.


libs/language-server/src/lib/validation/checks/constraint-definition.ts
```diff
@@ -45,4 +57,42 @@ export function checkConstraintExpression(
...
+      } else if (isReferenceLiteral(exp)) {
+        if (valueTypePropertyName === undefined) {
+          return undefined;
+        }
+        return valueTypePropertyName === exp.value.ref?.name ? true : undefined;
+      } else if (isFreeVariableLiteral(exp)) {
+        assertUnreachable(exp);
+      }
+      return undefined;
+    }) ?? false;
```
georg-schwarz, line 84: It might make sense to extract. Returning a boolean here threw me off a little :)

georg-schwarz (approved):
Nicely done! Left two things to potentially fix before merging

Issue #527, opened 2024-02-09
[BUG] JavaScript heap out of memory


kreisligaspieler:
## Steps to reproduce
Use following pipeline:
```
pipeline TestPipeline {
Extractor -> TextFileInterpreter->RangeSelector->CSVInterpreter->ColumnDeleter->TableInterpreter->Loader;
block Extractor oftype HttpExtractor {
            url: "https://derwebmaster.eu/wp-content/uploads/2024/02/star0000-1.csv";}
block TextFileInterpreter oftype TextFileInterpreter {encoding: "utf8";}
block RangeSelector oftype TextRangeSelector {lineFrom: 1;}
block CSVInterpreter oftype CSVInterpreter {enclosing: '';delimiter: ",";}
block ColumnDeleter oftype ColumnDeleter {delete: [];}
block TableInterpreter oftype TableInterpreter {
            header: false;
            columns: ["col1" oftype integer,"col2" oftype integer,"col3" oftype integer,"col4" oftype integer,"col5" oftype integer,"col6" oftype decimal,"col7" oftype integer,"col8" oftype integer,"col9" oftype decimal,"col10" oftype integer,"col11" oftype decimal,"col12" oftype integer];}
block Loader oftype SQLiteLoader {table: "tDbNEGvkAEdn"; file: "test.sqlite";}
}
```

Original source of the csv: https://sdm.lbl.gov/fastbit/data/samples.html   

## Description
- Expected:  Pipeline should download the file and store the content in a database
- Actual: FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory

**Debug:**
[TestPipeline] Overview:
	Blocks (7 blocks with 1 pipes):
	 -> Extractor (HttpExtractor)
		 -> TextFileInterpreter (TextFileInterpreter)
			 -> RangeSelector (TextRangeSelector)
				 -> CSVInterpreter (CSVInterpreter)
					 -> ColumnDeleter (ColumnDeleter)
						 -> TableInterpreter (TableInterpreter)
							 -> Loader (SQLiteLoader)

	[Extractor] Fetching raw data from https://derwebmaster.eu/wp-content/uploads/2024/02/star0000-1.csv

**Console output:** 
<--- Last few GCs --->

[...]

PR #669, opened 2025-07-03
 [FEATURE] Report download progress

TungstnBallon:
This PR adds a simple terminal output to `HTTPExtractor` that reports the current download progress.

e.g.
```
        [HTTPExtractor] Downloaded 35MB out of 102MB -- 34.3%
```

This was tested with the pipline from #527

### DIFF
--- a/libs/extensions/std/exec/src/http-extractor-executor.ts
+++ b/libs/extensions/std/exec/src/http-extractor-executor.ts
@@ -34,6 +34,47 @@ const { http, https } = followRedirects;
 
 type HttpGetFunction = typeof http.get;
 
+class DownloadProgress {
+  private downloadedBytes = 0;
+  private firstPrint = true;
+
+  constructor(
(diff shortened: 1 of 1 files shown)


libs/extensions/std/exec/src/http-extractor-executor.ts
```diff
@@ -34,6 +34,47 @@ const { http, https } = followRedirects;
...
+  private downloadedBytes = 0;
+  private firstPrint = true;
+
+  constructor(
+    private readonly logger: R.Logger,
+    private readonly totalBytes?: number,
+  ) {}
+
+  private formatBytes(bytes: number): string {
+    if (bytes / 1_000_000_000 >= 1) {
```
georg-schwarz, line 47: Interesting, the `_` notation, didn't know about it

georg-schwarz (approved):

PR #670, opened 2025-07-10
[RFC] 0017 error values

TungstnBallon:
This RFC introduces a rudimentary form of error handling.
Suggestions welcome!

### DIFF
--- a/rfc/0017-error-values/README.md
+++ b/rfc/0017-error-values/README.md
@@ -51,6 +51,9 @@ This RFC introduces two error values:
 This distinction is made in order to allow both users and the interpreter more
 fine grained control.
 
+For now, these values are valid for all types (see
+[Possible Future Changes/Enhancements](#possible-future-changesenhancements))
+
 ### invalid
 
 This error's primary use case is to represent an erroneous calculation result
--- a/rfc/0017-error-values/README.md
+++ b/rfc/0017-error-values/README.md
@@ -9,7 +9,7 @@ SPDX-License-Identifier: AGPL-3.0-only
 | | |
 |---|---|
 | Feature Tag | `error-values` |
-| Status | `DRAFT` | <!-- Possible values: DRAFT, DISCUSSION, ACCEPTED, REJECTED -->
+| Status | `ACCEPTED` | <!-- Possible values: DRAFT, DISCUSSION, ACCEPTED, REJECTED -->
 | Responsible | `tungstnballon` |
 <!--
   Status Overview:
--- a/rfc/0017-error-values/README.md
+++ b/rfc/0017-error-values/README.md
@@ -49,8 +49,10 @@ fine grained control.
 
 For now, these values are valid for the valuetypes `text`, `boolean`, `integer` and `decimal` (see
 [Possible Future Changes/Enhancements](#possible-future-changesenhancements)).
-However, the user is always notified (perhaps with a log message), when such
-a values occurs.
+
+When a value becomes `invalid` or `missing` for the first time, the user is
+notified with a log message. This message must contain the reason for the
+error's occurrence.
 
 This RFC also introduces two keywords `invalid` and `missing`, which represent
 the corresponding error value.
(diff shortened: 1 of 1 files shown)


georg-schwarz:
Congrats to your first submitted RFC! :tada: 

Here some feedback from my side:
1. The term "error handling" is somewhat ambiguous, since it might also expand to data sources being unreachable or some other issues. So I'd encourage speaking of "invalid and missing value handling".
2. Since two additional values are introduced, we should probably also speak about how they are represented in the language. As keywords? How do existing language elements interact with it, such as setting a value as invalid or missing?
3. IMO the level of abstraction is drawn nicely with the two values.

rfc/0017-error-values/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+- One single `ERROR`. 
+- More fine-grained errors. E.g `ParsingError`, `DivisionByZeroError`,
+`EmtyCellError`
+
+## Possible Future Changes/Enhancements
+
+- Allow the user to define the interpreter's behavior in case of an error. This
+  could mean replacing missing values, or crashing in case of an erroneous
+calculation.
```
georg-schwarz, line 85: Addition:
- Type unions to express a value can be "number or missing or invalid".

georg-schwarz, line 85: Addition:
- Preserve in the value the reason and the location why it became invalid (trace to where it happened)

TungstnBallon:
I've added a section describing how operators interact with the new values. Currently, the errors are 'viral', any operation with an error as a parameter also returns an error.

Some operations may be able to do this differently though, for example:
```
tansform tr {
    from s1 oftype text;
    from s2 oftype text;
    to out oftype boolean;
    out: s1 matches s2;
}
```
Assuming `s1` is a correct value and  `s2` is `missing`, this RFC specifies that the expression evaluates to `missing`.
However, `false` makes just as much sense to me.

The latter behavior may make it harder for users to notice errors.

What are your thoughts? @georg-schwarz @rhazn

rhazn:
I think making them viral is the better choice for now, especially because of the reason you mentioned (silently hiding missing values).

I see you updated the RFC with a sentence that these values will be valid for all value types for now. Before you set this to be ready for review, I'll just mention that I am unsure about that. I think it would be nice to at least have the interpreter log very visible errors when missing values or especially errors are encountered in this case.

rfc/0017-error-values/README.md
```diff
@@ -0,0 +1,181 @@
...
+  - REJECTED: The RFC was rejected. If another revision emerges, switch to status DRAFT.
+-->
+
+## Summary
+
+This RFC introduces the concept of invalid or missing values to the Jayvee interpreter.
+Specifically, it defines two new special values that the Jayvee interpreter must be able to handle.
+
+> [!NOTE]
+> This RFC does not define any changes to the Jayvee grammar
```
georg-schwarz, line 28: I think this is not true anymore.

rfc/0017-error-values/README.md
```diff
@@ -0,0 +1,181 @@
...
+  /*
+  | column | result  |
+  |--------|---------|
+  | 3      | invalid |
+  */
+  // `invalid` is replaced with `NULL`
+  block ToSQLite oftype SQLiteLoader {
+    table: "Table";
+    file: "data.sqlite";
+  }
```
georg-schwarz, line 150: I am unsure how we should handle the two introduced error values.
Until now, we cannot express something like "number or invalid" in table interpreters.
Should we assume every value can be x, invalid, or missing? Or proceed as we do right now and just remove every row that has invalid or missing values?

@rhazn thoughts?

rhazn: Invalid or missing are correct values for any value type according to this RFC so they will not be removed and instead just be saved as NULL.

Not sure if I have a strong opinion before the RFC is finished, so I'd wait for it to be out of draft before reviewing.

georg-schwarz (approved):
+1 for Philip's suggestions. Don't forget to change the status of the RFC before merging ;-)

PR #673, opened 2025-08-04
[RFC0017] Add `invalid` and `missing`error values

TungstnBallon:
Depends on #674.

This PR implements RFC0017, accepted in #670.

Also included is a restructuring of `PostgresLoaderExecutor` with a more granular `try/catch`.

Reason for the draft status: The RFC states that a log message must be emitted every time an error is assigned to a variable. I still need to make sure that is actually the case in the implementation

Edit: sorry for the large diff

### DIFF
--- a/libs/execution/src/lib/types/io-types/table.ts
+++ b/libs/execution/src/lib/types/io-types/table.ts
@@ -8,8 +8,8 @@ import { strict as assert } from 'assert';
 import {
   ERROR_TYPEGUARD,
   IOType,
-  type InternalErrorRepresentation,
-  type InternalValueRepresentation,
+  type InternalErrorValueRepresentation,
+  type InternalValidValueRepresentation,
   type TextValuetype,
   type ValueType,
   cloneInternalValue,
@@ -24,15 +24,15 @@ import {
 } from './io-type-implementation';
 
 export interface TableColumn<
-  T extends InternalValueRepresentation = InternalValueRepresentation,
+  T extends InternalValidValueRepresentation = InternalValidValueRepresentation,
 > {
-  values: (T | InternalErrorRepresentation)[];
+  values: (T | InternalErrorValueRepresentation)[];
   valueType: ValueType;
 }
 
 export type TableRow = Record<
   string,
-  InternalValueRepresentation | InternalErrorRepresentation
+  InternalValidValueRepresentation | InternalErrorValueRepresentation
 >;
 
 /**
@@ -80,7 +80,7 @@ export class Table implements IOTypeImplementation<IOType.TABLE> {
 
       assert(
         ERROR_TYPEGUARD(value) ||
-          column.valueType.isInternalValueRepresentation(value),
+          column.valueType.isInternalValidValueRepresentation(value),
       );
       column.values.push(value);
     });
@@ -110,7 +110,10 @@ export class Table implements IOTypeImplementation<IOType.TABLE> {
 
   getRow(
     rowId: number,
-  ): Map<string, InternalValueRepresentation | InternalErrorRepresentation> {
+  ): Map<
+    string,
+    InternalValidValueRepresentation | InternalErrorValueRepresentation
+  > {
     const numberOfRows = this.getNumberOfRows();
     if (rowId >= numberOfRows) {
       throw new Error(
@@ -120,7 +123,7 @@ export class Table implements IOTypeImplementation<IOType.TABLE> {
 
     const row = new Map<
       string,
-      InternalValueRepresentation | InternalErrorRepresentation
+      InternalValidValueRepresentation | InternalErrorValueRepresentation
     >();
     [...this.columns.entries()].forEach(([columnName, column]) => {
       const value = column.values[rowId];
--- a/libs/execution/src/lib/types/io-types/table.ts
+++ b/libs/execution/src/lib/types/io-types/table.ts
@@ -6,7 +6,9 @@
 import { strict as assert } from 'assert';
 
 import {
+  ERROR_TYPEGUARD,
   IOType,
+  type InternalErrorRepresentation,
   type InternalValueRepresentation,
   type TextValuetype,
   type ValueType,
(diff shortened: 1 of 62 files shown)


libs/execution/src/lib/types/io-types/table.ts
```diff
@@ -20,14 +22,34 @@ import {
...
+  clone(): TableColumn<T> {
+    const cloned: (T | InternalErrorRepresentation)[] = [];
+    for (const value of this.values) {
+      if (ERROR_TYPEGUARD(value)) {
+        cloned.push(value.clone());
+      } else {
+        // WARNING: structuredClone does not support all types!
+        // See
+        // https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#supported_types
+        cloned.push(structuredClone(value));
```
georg-schwarz, line 42: Is this an open TODO?

libs/interpreter-lib/src/validation-checks/runtime-parameter-literal.ts
```diff
@@ -83,18 +84,10 @@ function checkRuntimeParameterValueParsing(
...
-      props.evaluationContext.runtimeParameterProvider.getRawValue(
-        runtimeParameterName,
-      );
-    props.validationContext.accept(
-      'error',
-      `Unable to parse the value "${
-        rawValue ?? ''
-      }" as ${valueType.getName()}.`,
-      { node: runtimeParameter },
-    );
```
georg-schwarz, line 97: Why is this code removed? Did we introduce changes to the runtime parameter handing with this PR?

libs/language-server/src/lib/ast/expressions/evaluators/parse-operators-evaluators.ts
```diff
@@ -42,13 +38,8 @@ export class AsIntegerOperatorEvaluator extends DefaultUnaryOperatorEvaluator<
...
   }
-  override doEvaluate(operandValue: string): number {
-    const int =
-      this.valueTypeProvider.Primitives.Integer.fromString(operandValue);
-    if (int === undefined) {
-      throw new Error(`Could not parse "${operandValue}" into an Integer`);
-    }
-    return int;
+  override doEvaluate(operandValue: string): number | InvalidError {
+    return this.valueTypeProvider.Primitives.Integer.fromString(operandValue);
```
georg-schwarz, line 42: Nice to see that things actually get easier at a few places! :tada:

libs/language-server/src/lib/ast/expressions/internal-value-representation.ts
```diff
@@ -2,6 +2,11 @@
 //
 // SPDX-License-Identifier: AGPL-3.0-only
 
+// eslint-disable-next-line unicorn/prefer-node-protocol
+import assert from 'assert';
```
georg-schwarz, line 6:
```suggestion
import assert from 'node:assert';
```

libs/language-server/src/lib/ast/expressions/internal-value-representation.ts
```diff
@@ -14,7 +19,50 @@ import {
...
+  }
+
+  override toString(): string {
+    return `${this.name}: ${this.message}`;
+  }
+
+  abstract clone(): JayveeError;
+}
+
+export class InvalidError extends JayveeError {
```
georg-schwarz, line 38: I think we should call it `InvalidValue` instead. But might be personal preference, opinions?

libs/language-server/src/lib/ast/expressions/internal-value-representation.ts
```diff
@@ -14,7 +19,50 @@ import {
...
+      this.stack = stack;
+    }
+  }
+
+  override clone(): InvalidError {
+    return new InvalidError(this.message, this.stack);
+  }
+}
+
+export class MissingError extends JayveeError {
```
georg-schwarz, line 53: I think we should call it `MissingValue` instead. But might be personal preference, opinions?

libs/language-server/src/lib/ast/expressions/internal-value-representation.ts
```diff
@@ -14,7 +19,50 @@ import {
   isTransformDefinition,
   isValuetypeAssignment,
 } from '../generated/ast';
-import type { WrapperFactoryProvider } from '../wrappers';
+import { type WrapperFactoryProvider } from '../wrappers';
+
+abstract class JayveeError extends Error {
```
georg-schwarz, line 24: I think we should call it `ErroneousValue` instead. But might be personal preference, opinions?

libs/language-server/src/lib/ast/expressions/typeguards.ts
```diff
@@ -55,6 +58,16 @@ export const COLLECTION_TYPEGUARD: InternalValueRepresentationTypeguard<
...
+
+export const MISSING_TYPEGUARD = (value: unknown): value is MissingError => {
+  return value instanceof Error && value.name === 'MissingError';
+};
+
+export const ERROR_TYPEGUARD = (
+  value: unknown,
+): value is InternalErrorRepresentation => {
+  return INVALID_TYPEGUARD(value) || MISSING_TYPEGUARD(value);
+};
```
georg-schwarz, line 73: Would use classic typeguard function naming `is<TypeName>`

georg-schwarz (commented):
Very nice progress! Left some comments to discuss and/or address ;-)

libs/language-server/src/lib/ast/expressions/typeguards.ts
georg-schwarz: Seems to be used similarly on other occasions as well... So align with what is there. Potentially, refactor in the future.

libs/execution/src/lib/types/io-types/table.ts
TungstnBallon: I have decided to tackle this problem in #674

PR #675, opened 2025-08-26
[CHORE] Update some outdated dependencies

TungstnBallon:
This PR upgrades some of the outdated dependencies to their latest versions:
- node types
- nx
- vitest
- prettier
- eslint


libs/execution/src/lib/transforms/transform-executor.spec.ts
```diff
@@ -36,7 +36,7 @@ import { type Table, type TableColumn } from '../types/io-types/table';
 import { type PortDetails, TransformExecutor } from './transform-executor';
 
 function expectNoErrorsInColumn(column: TableColumn) {
-  expect(column.values.every((value) => ERROR_TYPEGUARD(value)));
+  column.values.every((value) => expect(value).not.toSatisfy(ERROR_TYPEGUARD));
```
georg-schwarz, line 39: Is this the opposite behavior from before?

TungstnBallon: Yes, this was actually a bug I found through an eslint warning.

PR #677, opened 2025-09-11
RFC0018: Multi-attribute value types

jrentlez:
This RFC allows value types to have multiple attributes.
~~Value types are used to define tables, replacing the existing syntax.~~

### DIFF
--- a/rfc/0018-multi-attribute-value-types/README.md
+++ b/rfc/0018-multi-attribute-value-types/README.md
@@ -9,7 +9,7 @@ SPDX-License-Identifier: AGPL-3.0-only
 | | |
 |---|---|
 | Feature Tag | `multi-attribute value types` |
-| Status | `DISCUSSION` | <!-- Possible values: DRAFT, DISCUSSION, ACCEPTED, REJECTED -->
+| Status | `ACCEPTED` | <!-- Possible values: DRAFT, DISCUSSION, ACCEPTED, REJECTED -->
 | Responsible | `jrentlez` |
 <!-- 
   Status Overview:


georg-schwarz (commented):
I like the approach. @rhazn thoughts?

A little bit connected is how we will handle parsing (e.g., to handle renaming of columns better). But I'd be in favor to force programmers to give better names to coumns if they have whitespaces etc. to not address them with string indeces but properties.

rfc/0018-multi-attribute-value-types/README.md
```diff
@@ -0,0 +1,153 @@
...
+valuetype Coordinate2D {
+  property x oftype decimal;
+  property y oftype decimal;
+
+  constraint isOn45DegreeLine: x == y;
+}
+
+block toTable oftype TableInterpreter {
+  header: true;
+  type: Coordinate2D;
```
georg-schwarz, line 119: I'd call it `schema`, but I'm other to other suggestions :D

rfc/0018-multi-attribute-value-types/README.md
```diff
@@ -0,0 +1,162 @@
...
+Conceptually, a value type defines a table. The instance of that value type is a
+row inside the table.
+
+Nested value types are allowed:
+```jayvee
+valuetype Coordinate2D {
+  property x oftype decimal;
+  property y oftype decimal;
+}
+valuetype Circlde {
```
georg-schwarz, line 97:
```suggestion
valuetype Circle {
```

rfc/0018-multi-attribute-value-types/README.md
```diff
@@ -0,0 +1,162 @@
...
+  schema: {
+    property x oftype decimal;
+    property y oftype decimal;
+  }
+}
+```
+
+One limitation with the new syntax is that column names have to match the jayvee
+identifier regex `/[_a-zA-Z][\w_]*/`, which excludes many expected names (e.g.
+with spaces).<br>
```
rhazn, line 82: Those names are valid names for sinks as well though (e.g. for CSV sinks, but also for SQLite I think). Is the intended solution then that Jayvee is just unable to produce those headers in sinks?

rfc/0018-multi-attribute-value-types/README.md
```diff
@@ -0,0 +1,162 @@
...
+The intended solution for this is a new composite block in the standard library,
+combining `CellWriter`, to rename the columns into a compatible format and
+`TableInterpreter`.<br>
+This follows the jayvee design principle of *libraries over language features*.
+
+Conceptually, a value type defines a table. The instance of that value type is a
+row inside the table.
+
+Nested value types are allowed:
```
rhazn, line 91: I see some of my questions are answered here, this section should be moved up before the explanation of schemas/the schemas should not be discussed here.

rfc/0018-multi-attribute-value-types/README.md
```diff
@@ -0,0 +1,162 @@
...
+  property x oftype decimal;
+  property y oftype decimal;
+}
+valuetype Circle {
+  property center oftype Coordinate2D;
+  property radius oftype decimal;
+}
+```
+
+At first, the jayvee interpreter will only implement nested value types, where the
```
rhazn, line 103: Are we using some `.value` magic naming? Or is the reference still by the name of the property? What does "implement" mean, e.g. what is the intended behavior otherwise? If it is "it crashes", this should be part of the RFC.

rfc/0018-multi-attribute-value-types/README.md
```diff
@@ -0,0 +1,162 @@
...
+
+- Don't introduce multi attribute value types.
+
+## Possible Future Changes/Enhancements
+
+- Implement nested value types with multiple attributes in the interpreter. This
+  probably requires dynamically creating a new table and putting foreign keys
+  into the cell the value should be.
+
+- Define to inline the valuetype into `TableInterpreter` to reduce verbosity.
```
rhazn, line 161: I can not follow this sentence.

georg-schwarz (approved):

PR #685, opened 2025-11-10
rfc(0020): table-schema-value-type

jrentlez:
See README.md

### DIFF
--- a/rfc/0020-table-value-types/README.md
+++ b/rfc/0020-table-value-types/README.md
@@ -9,7 +9,7 @@ SPDX-License-Identifier: AGPL-3.0-only
 | | |
 |---|---|
 | Feature Tag | `table-schema-value-types` |
-| Status | `DRAFT` |
+| Status | `ACCEPTED` |
 | Responsible | `jrentlez` |
 <!-- 
   Status Overview:


rfc/0020-table-value-types/README.md
```diff
@@ -0,0 +1,197 @@
...
+## Summary
+
+This RFC changes the way sheets are parsed into tables, using a multi property
+value type as the table schema definition.
+
+## Motivation
+
+- Allows users to define how a sheet is parsed into a table.
+- Allows users to reuse table schema definitions.
+- Remove special syntax regarding table definitions.
```
georg-schwarz: Not a motivation; rather focus on what the current problem is.. Are we not able to express something, is the syntax not consistent, ...

rfc/0020-table-value-types/README.md
```diff
@@ -0,0 +1,197 @@
...
+  rowParsedWith: <transform>;
+}
+```
+
+Because `<transform>` parses rows, it needs to conform to:
+- only one input of type `SheetRow`.
+- only one output with the same value type as in `<block>`'s `columns` property.
+- must use a row-creation expression (see below)
+
+### `SheetRow`
```
georg-schwarz: Foreshadow a little bit at the beginning that we need to introduce multiple concepts (otherwise this seems to be random as if it should belong into another RFC). Have a list of things this RFC introduces at the start:
- A new syntax to define columns of a table via a valuetype
- SheetRow builtin value type with new cell/field access syntax
- - ...

PR #690, opened 2026-02-19
RFC 0021: Implementation

jrentlez:
This PR implements RFC 0021 (#689)


georg-schwarz (approved):
Awesome! I'll trigger copilt but good to go from my side!

PR #691, opened 2026-03-05
fix: CI failure on main

jrentlez:
This PR fixes the [CI failure on main](https://github.com/jvalue/jayvee/actions/runs/22309652396/job/64537936406).

The fundamental problem is that `SheetRow` does not override `getUserDoc()`, which causes `nx run docs-generator:start` to fail. This is only noticeable in the CI, because existing repos have a left-over `builtin-value-types.md` in the source tree, which is used by `nx build docs` instead.

However, while working on this I discovered another issue:
`nx run docs-generator:start` stops before ever getting to the above failure, without any error log or error code. The exact stop location is very inconsistent. Unfortunately, I got on the wrong track here and spent quite some time looking into `langium` and `chevrotain` source code, to try and figure out where exactly this crash originates.
Long story short, this happens because nx stops the underlying `node` process after a timeout, attempts to restart it, and then continues execution without any hint to the user. 

See the commit messages for more details.

I decided to fix both issues in one PR, because fixing only one issue would leave code we know to be erroneous in main.

### DIFF
--- a/apps/docs-generator/project.json
+++ b/apps/docs-generator/project.json
@@ -13,11 +13,11 @@
       }
     },
     "start": {
-      "executor": "@nx/js:node",
-      "dependsOn": ["generate-language-server"],
+      "executor": "nx:run-commands",
+      "dependsOn": ["generate-language-server", "build"],
       "options": {
-        "watch": false,
-        "buildTarget": "docs-generator:build"
+        "commands": ["node dist/apps/docs-generator/main.js"],
+        "parallel": false
       }
     },
     "generate-language-server": {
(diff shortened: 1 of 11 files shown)


apps/docs-generator/project.json
```diff
@@ -13,11 +13,11 @@
       }
     },
     "start": {
-      "executor": "@nx/js:node",
-      "dependsOn": ["generate-language-server"],
+      "executor": "nx:run-commands",
```
georg-schwarz, line 16: Could you please add a comment here and say that this did not work with `@nx/js:node` and link the PR?

libs/language-server/src/lib/ast/wrappers/value-type/primitive/primitive-value-type.ts
```diff
@@ -33,10 +34,39 @@ export abstract class PrimitiveValueType<
...
+}
+block ExampleTableInterpreter oftype TableInterpreter {
+  header: true;
+  columns: TableSchema;
+  parseWith: tableRowParser;
+}`.trim(),
+        description:
+          `A block of type \`TableInterpreter\` that interprets data in the column "columnName" as \`${name}\`.
+              `.trim(),
+      },
```
georg-schwarz, line 66: Would be cool to have a real example / close to a real use-case and not a dummy one

georg-schwarz (approved):
LGTM, nicely done!

